PR #166, opened 2025-11-10
Use haiku.rag research graph with the agent factory.

ggozad:
This is a first-attempt to wrap the haiku.rag research graph within an agent using the agent-factory support.

This is WIP, the agent will:

1. Extract the user's question from the message history (last ModelRequest with UserPromptPart)
2. Get RAG database path from the room's search_documents tool config
3. Build research graph and execute it via stream_research_graph():
  - Creates ResearchContext with the original question
  - Creates ResearchState with config from extra_config (max_iterations, confidence_threshold, max_concurrency)
  - Creates ResearchDeps with the RAG client
4. Stream events from the research graph:
  - "log" events: Progress messages (sub-questions, insights, gaps, confidence) streamed as TextParts
  - "report" events: Final ResearchReport, formatted as markdown with sections (executive summary, findings, conclusions, limitations, recommendations, sources)
  - "error" events: Errors during research → streamed as error messages

So basically I use `stream_research_graph ` which I had written to output to the console events. This should be ideally done using `agui` within `haiku.rag` instead, I just wanted to get something out as a quick & dirty first draft on how to use the pydantic AI graphs.

Alternatively, one could do without the graph altogether and use a multi-agent with tools. This would probably be simpler but more verbose and I am guessing more fragile.


ggozad:
I think we can close this one, it's out of date with what has been happening. 
@tseaver would it be possible to have a look at the [ag-ui-example](https://github.com/ggozad/haiku.rag/tree/main/examples/ag-ui-research/backend)? I am not sure if the pattern I follow there is how you want to integrate graphs/subagents with soliplex, but in there you can find:
1. A rather simple agent in `agent.py` that basically runs the graph as a tool. In the graph's dependencies it passes a custom `AGUIEmitter` so the graph writes events to it while it executes.
2. In `main.py` you will find the `starlette` route that streams the event stream.

I think some thinking needs to go into this to see how it could be used in soliplex generically so that we could in principle use any agent with graphs or other subagents as tools. That's the general idea, but @tseaver can probably improve on this and come up with something better.

tseaver:
I believe this is superseded by #205.

PR #192, opened 2025-11-18
docs: show use of the TUI client in the README

### LINKED ISSUE #190: Expose TUI in the README
... and make it speak AG-UI by default.

/cc @runyaga

tseaver:
Change it's default for connecting via AG-UI to True (and rename the option to '--agui' / '-no-agui').

Closes #190.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,18 @@ haiku-rag --config example/haiku.rag.yaml \
 
 See: `docs/rag.md` for more options.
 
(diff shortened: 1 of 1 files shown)


tseaver:
@runyaga 
> what about showing a curl command or --list command so one could list rooms? I always forget the incantation

Doesn't this work for you?
```bash
soliplex-cli list-rooms example/minimal.yaml
```

PR #194, opened 2025-11-18
!chore: disuse 'DEFAULT_AGENT_MODEL' env setting

### LINKED ISSUE #191: get rid of DEFAULT_AGENT_MODEL legacy now we have `agent_configs`?
now that `agent_configs:` was introduced what purpose is global "DEFAULT_AGENT_MODEL"?

What about a required default that blows up if its not defined?

```
agent_configs:
      - id: "default"
         model_name: "qwen3:latest"
```

tseaver:
NOTE:  this change potentially breaks users' installation configs.

Agent configurations must now configure their 'model_name', either directly or via their template.

Closes #191.


tseaver:
I think we'll hold off on this one, and instead deprecate the `DEFAULT_AGENT_MODEL` for a bit, cleaning up our `example` configs so that they use agent config templates with explicit models.  We can do the actual removal of the default after a bit.

PR #217, opened 2025-11-25
Agui framework

91jaeminjo:


src/flutter/lib/infrastructure/quick_agui/text_message_buffer.dart
```diff
@@ -0,0 +1,24 @@
...
+  final StringBuffer messageId = StringBuffer();
+  final StringBuffer _buffer = StringBuffer();
+
+  TextMessageBuffer();
+
+  void start(String id) {
+    messageId.clear();
+    messageId.write(id);
+    _buffer.clear();
+  }
```
svarlet: Do we need a start method? or could the constructor take care of this initialisation work?

PR #235, opened 2025-12-02
feat: optional Ollama/OpenAI model settings

### LINKED ISSUE #140: model settings: temp, top_k, etc
We have not encountered the need but I believe we will want to expose model settings such as
https://ai.pydantic.dev/api/settings/#pydantic_ai.settings.ModelSettings

tseaver:
Closes #140.


tseaver:
This PR isn't ready:  it allows you to enter model settings as part of agent config, but then doesn't actually do anything with them.

PR #254, opened 2025-12-04
FIX: Fix/agui toolcall

91jaeminjo:
Fixing client side tool call infinite loop.
Client side tool registry wasn't marking the tool call as complete after receiving the result.
Thus the client side kept making that tool call.

### DIFF
--- a/src/flutter/test/infrastructure/quick_agui/thread_test.dart
+++ b/src/flutter/test/infrastructure/quick_agui/thread_test.dart
@@ -451,6 +451,23 @@ void main() {
       );
 
       expect(result.content, equals('{"sum":3}'));
+
+      clientWillReceive(
+        client,
+        aRunWithEvents(threadId, runId, [
+          ag_ui.TextMessageChunkEvent(
+            messageId: 'msg-3',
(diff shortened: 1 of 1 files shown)


91jaeminjo:
When there is a pending client side tool call, the client makes those tool calls, and starts a new run with those tool call results.
Once those results are sent back, I needed to mark those tool calls as complete, but I forgot to.
So even after the assistant moves on with the next message, the client keeps performing those tool calls, and sending the results back to the assistant..

I have separate unit tests in the `ToolCallRegistry` that tests that marking the tool call as complete will take it out of pending tool call, but didn't have a test that tested all these in action..

tseaver:
> I have separate unit tests in the ToolCallRegistry that tests that marking the tool call as complete will take it out of pending tool call, but didn't have a test that tested all these in action..

Should this PR therefore add a test (or added assertion) that the completed tool calls get marked as complete?

91jaeminjo:
I added to the thread_test.dart to verify that after sending the tool result in the 2nd run, that there is no pending client side tool call.
I verified that this test fails before the fix, and now passes with the fix.

PR #291, opened 2025-12-12
haiku.rag 0.20.2

ggozad:
Updates for haiku.rag 0.20.1
Fixed some tests that did not pass on macos due to temp file resolution.

@tseaver please check this thoroughly, I changed what I remembered was there and had a good look, but haven't run anything other than tests to verify.

### DIFF
--- a/example/rooms/haiku/haiku.rag.yaml
+++ b/example/rooms/haiku/haiku.rag.yaml
@@ -7,18 +7,17 @@
 #
 # Uncomment to enable reranking for the 'search_documents' tool
 #
+# See: https://ggozad.github.io/haiku.rag/configuration/providers/#reranking-providers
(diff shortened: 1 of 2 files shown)


example/haiku.rag.yaml
```diff
@@ -1,13 +1,11 @@
...
 
-# Configure 'production' for standard CLI logging, 'development' for Logfire
 environment: development
 
 storage:
-  data_dir: ''
-  disable_autocreate: true
-  vacuum_retention_seconds: 60
+  data_dir: ""
+  vacuum_retention_seconds: 86400
```
tseaver: Is that the `haiku-rag` slipping in by default, or do you recommend that we actually increase this value?

PR #412, opened 2026-01-07
feat(ci): add mypy type checking (advisory mode)

runyaga:
## Summary
- Add mypy to dev dependencies for type checking
- Add mypy step to CI workflow in advisory mode
- Configure mypy with relaxed settings for gradual adoption

## Changes
- **pyproject.toml**: Added mypy to dev dependencies, configured `[tool.mypy]` section
- **python-test.yaml**: Added mypy step with `continue-on-error: true`

## Configuration Details
- Excludes `tests/` and `src/soliplex/tui/` directories
- Ignores missing imports for third-party libraries without stubs
- Disables strict untyped checks for gradual adoption

## Current Status
260 type errors to be fixed incrementally. Mypy runs in advisory mode
so CI won't fail, but errors are visible in the workflow output.

## Test plan
- [x] mypy runs locally
- [ ] CI passes (advisory mode)

**Note:** This PR is based on `fix/ci-improvements` branch (PR #411) and should be merged after that PR.

### DIFF
--- a/.github/workflows/python-test.yaml
+++ b/.github/workflows/python-test.yaml
@@ -63,6 +63,10 @@ jobs:
       - name: Run ruff linter
         run: ruff check
 
+      - name: Run mypy type checker
+        run: mypy
+        continue-on-error: true  # Advisory for now, fix types gradually
+
       - name: Security scan dependencies
         run: pip-audit --strict --progress-spinner=off
         continue-on-error: true  # Advisory for now, can make strict later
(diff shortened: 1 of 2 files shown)


tseaver (commented):
Before merging, we need to figure out why this PR didn't trigger the `python-test.yaml` workflow:  otherwise, we don't know that the check it adds actually works.

tseaver:
Also, I'm radically opposed to adding such a check without making the developer who proposes it do the work to make it pass.  Either the check is worthwhile (a debatable proposition), or it is not:  "there is no try."

PR #445, opened 2026-01-15
feature: add pull-models command to soliplex-cli

mcdonc:


tests/unit/test_secrets.py
```diff
@@ -116,6 +116,16 @@ def test_get_subprocess_secret(command, args, expectation, expected):
         assert found == expected
 
 
+@mock.patch("subprocess.check_output")
+def test_get_subprocess_secret_empty_output(sco):
+    sco.return_value = ""
+
+    source = config.SubprocessSecretSource(SECRET_NAME, "some_cmd", ())
+
+    with SubprocessError:
```
tseaver: I'd definitely prefer to show raising the specific error here, e.g.:
```
    with secrets.SecretSubprocessError:
```

PR #530, opened 2026-01-30
tui improvements, haiku.rag 0.27.2

ggozad:
Here's the gist:

1. Update TUI to construct initial AGUI state from features. 
2. Add `haiku.rag.chat` feature in the `chat` room
3. Add a modal to the TUI to observe the AGUI state (ctrl+s)
4. Logfire setup now also looks at .env
5. haiku.rag updated to latest.

### DIFF
--- a/src/soliplex/tui/main.py
+++ b/src/soliplex/tui/main.py
@@ -338,6 +338,49 @@ class RunEventWidget(t_widget.Widget):
         yield t_widgets.Static(self.event_content)
 
 
+class StateViewModal(t_screen.ModalScreen):
+    BINDINGS = [
+        t_binding.Binding("escape", "dismiss(None)", "Close"),
+    ]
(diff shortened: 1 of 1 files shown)


example/minimal.yaml
```diff
@@ -128,6 +128,7 @@ secrets:
 
 environment:
   - "OLLAMA_BASE_URL"
+  - "LOGFIRE_TOKEN"
```
tseaver: This bit is problematic:  `LOGFIRE_TOKEN` is a secret, and not a regular environment variable.  The `minimal.yaml` config doesn't require *any* secrets (by design).

I'd remove this bit, and the bits below which change how `installation.apply_logfire_configuration` works, and instead update the docs to emphasize that secrets need to be in the actual OS environment, and not in a `.env` file.

PR #617, opened 2026-02-16
Integrate haiku.rag 0.30.1 toolset architecture into Soliplex

ggozad:
Replace custom RAG tools (search_documents, research_report,
ask_with_rich_citations) with haiku.rag's composable toolset system
via ChatAgentWrapper and ChatAgentConfig.

Key changes:
- ChatAgentWrapper uses build_chat_toolkit/create_chat_agent with
  ToolContextCache to persist session context across requests,
  enabling background summarization
- AgentDependencies gains thread_id field for context cache keying
- Remove SearchDocumentsToolConfig, RAGResearchToolConfig,
  AskWithRichCitationsToolConfig and associated tool functions
- Simplify agui/features.py to just HAIKU_CHAT_FEATURE
- Add features/preamble fields to ChatAgentConfig
- Delete rooms that depended on removed tools (ask_soliplex, haiku,
  research); keep all non-haiku-rag rooms
- Fix TUI STATE_SNAPSHOT event using wrong key (content → snapshot)
- Bump haiku.rag-slim >= 0.30.0

### DIFF
--- a/docs/config/agents.md
+++ b/docs/config/agents.md
@@ -51,7 +51,7 @@ agent:
         You are a knowledgeable assistant that helps users find information from a document knowledge base.
 
         Your process:
-        1. When a user asks a question, use the search_documents tool to find relevant information
+        1. When a user asks a question, find relevant information
         ...
 ```
 
--- a/docs/prerequisites.md
+++ b/docs/prerequisites.md
@@ -279,8 +279,8 @@ soliplex-cli list-rooms example/minimal.yaml
 ```
 
 You should see rooms like:
-- `ask_soliplex` - RAG-powered Soliplex documentation assistant
-- `haiku` - General purpose chat
+- `chat` - Conversational RAG with search and document retrieval
+- `search` - Search the knowledge base and answer questions
 - `joker` - Entertainment/joke generation
 - `faux` - Test room
 


example/minimal.yaml
tseaver: Same changes here as for the `installation.yaml` file.

src/soliplex/tui/main.py
```diff
@@ -1026,7 +1026,7 @@ def send_agui_prompt(self, prompt: str, response: Response) -> None:
 
                 elif chunk["type"] == "STATE_SNAPSHOT" and self.verbose:
                     response_content += (
-                        f"\n\n** state snapshot **\n\n{chunk['content']}\n\n"
+                        f"\n\n** state snapshot **\n\n{chunk['snapshot']}\n\n"
```
tseaver: Nice catch!  Is the `haiku_rag_chat` agent emitting snapshots now?  If so, how does it deal the the presence of other potential features in the state?

ggozad: There was a mixture of deltas with state snapshots. I thought it would be ok because the of the keyed state, but I now think it's safer to just use deltas. I will release 0.30.1 and pin to that one in soliplex.

PR #657, opened 2026-03-01
feat(rooms): add 16 M7 dedicated test rooms

runyaga:
## Summary
- Add 16 dedicated backend room configs for M7 integration tests
- Each room targets a specific agentic scenario (echo, tool-call, multi-tool, parallel, planner, critic, advocate, judge, etc.)
- All rooms use explicit model config (`gpt-oss:20b` on `workshop:11434`) instead of `template_id`

## Changes
- **`example/rooms/`**: 16 new `room_config.yaml` files — echo, tool-call, multi-tool, parallel, planner, critic, advocate, judge, classifier, dispatch, accumulator, fixer, reviewer, searcher, strict-tool, writer

## Test plan
- [x] Backend loads all 16 rooms (`curl /api/v1/rooms`)
- [x] All 20 M7 integration tests pass against these rooms
- [x] Existing M4/M5/M6 integration tests unaffected

Companion PR: soliplex/flutter (M7 integration tests)

### DIFF
--- a/example/rooms/accumulator/room_config.yaml
+++ b/example/rooms/accumulator/room_config.yaml
@@ -1,0 +1,10 @@
+id: "accumulator"
+name: "Accumulator Test"
+description: "Multi-turn context retention across tool calls"
+agent:
+  template_id: "default_chat"
+  system_prompt: |
+    You are a memory assistant. When the user gives you facts, acknowledge
+    them briefly. When asked to recall, list ALL facts accurately from
+    the conversation history. Do not make up facts.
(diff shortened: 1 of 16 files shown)


tseaver (changes requested):
Rather than jamming these rooms into the other example installations, can we put these new rooms under a separate directory, e.g. `example/m7-rooms/`, and add a top-level `installation-m7.yaml` which pulls them in?

Or maybe create an `m7-integration` top-level dir, with these rooms and its own `installation.yaml` file?

PR #681, opened 2026-03-09
fix: update DB for thread persistence/usage in background

### LINKED ISSUE #677: Split end-of-thread persistence out from streaming response completion
We need to commit the thread history / usage stats, etc., even in the presence of a `TCP_RST` ~~from the client~~ generated by client / network behavior.  In particular,  clients sometimes seem (from the server side) to disconnect abruptly / uncleanly:
- Clients may crash / disconnect before beginning to consume the SSE event stream.
- Clients may crash / disconnect partway through consuming the SSE event stream.

@runyaga the "Zombie LLM claim" in #680 seems invalid:  we need to record the completed run, even if the client disconnects:  for instance, when they reconnect, that run will be available in the thread history. Also, the end-of-thread callback is the only place where we can obtain the token usage data for the run.

Other sources of similar errors:

tseaver:
Drive LLM event stream, making DB updates in a task, dispatching events to the 'FastAPI.StreamingResponse' via a queue.

Closes #677.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "soliplex"
-version = "0.44dev0"
+version = "0.45dev0"
 description = "An AI-powered Retrieval-Augmented Generation (RAG) system with a modern web interface."
 authors = [{ name = "Enfold", email = "info@enfoldsystems.net" }]
 readme = "README.md"
--- a/src/soliplex/views/agui.py
+++ b/src/soliplex/views/agui.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-import functools
+import asyncio
 
 import fastapi
 import logfire
@@ -8,6 +8,7 @@ import pydantic_ai
 from ag_ui import core as agui_core
 from fastapi import responses
 from pydantic_ai.ui import ag_ui as ai_ag_ui
+from sqlalchemy.ext import asyncio as sqla_asyncio
 
 from soliplex import agui as agui_package
 from soliplex import authn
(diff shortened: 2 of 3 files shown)


mcdonc (approved):

runyaga:
## SSE Streaming Review — PR #681

PR #681 resolves two of the four issues identified during SSE streaming analysis (7 integration tests, 3 rounds of review with a Principal SRE persona).

### What this PR fixes

| Issue | Status | Notes |
|-------|--------|-------|
| DB persistence on TCP RST | **FIXED** | `save_thread_run_events()` creates its own `AsyncSession` independent of request lifecycle. Correct approach. |
| Anti-buffering headers | **FIXED** | `HEADERS_DO_NOT_BUFFER_SSE` with `Cache-Control: no-cache, no-store` and `X-Accel-Buffering: no`. Correct. |

### What remains unaddressed

#### C4: No heartbeat / keepalive

This is the most likely cause of the intermittent production hangs. During LLM thinking phases (can be 30-60s), zero bytes flow on the wire. Akamai GCDS and AWS ELB both enforce idle timeouts (typically 60s). When the timeout fires, the intermediary drops the TCP connection. The client sees a hang or abrupt close.

The standard SSE fix is to emit `: keepalive\n\n` (SSE comment) every 15s during idle periods. This is a no-op for clients but resets idle timers on every hop in the path (Client → Akamai → ELB → ALB → nginx → Uvicorn).

The current `stream_llm_events` does a bare `await event_queue.get()` with no timeout, so there is no opportunity to inject keepalives during idle periods.

#### C6: Task lifecycle — `asyncio.create_task()` without reference

`asyncio.create_task(drive_llm_stream(...))` does not store the returned `Task` object. Per the [Python docs](https://docs.python.org/3/library/asyncio-task.html#creating-tasks):

> **Important:** Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks.

Additionally, if `drive_llm_stream` raises an unhandled exception, it will surface as a `Task exception was never retrieved` warning in logs rather than propagating to the request handler. Storing the reference and awaiting or shielding it would make error handling explicit.

#### C7: `stream_llm_events` docstring vs. implementation

The docstring states "Stop if client disconnects" but the implementation has no disconnect detection mechanism:

```python
async def stream_llm_events(event_queue: asyncio.Queue):
    """Read/yield events from queue

    Stop if client disconnects.
    """
    while True:
        event = await event_queue.get()
        if event is None:
            break
        yield event
```

When a client disconnects, Starlette stops pulling from the generator but does not raise `CancelledError` into it (confirmed by integration test T2 and T7 in our test suite). The generator blocks on `queue.get()` until `drive_llm_stream` pushes the `None` sentinel after LLM completion. This means the generator lifetime is coupled to LLM completion, not client presence.

For disconnect detection, `request.is_disconnected()` polling via `asyncio.wait(timeout=N)` on the queue get is required.

#### Wasted tokens on early disconnect

When a client disconnects mid-stream, `drive_llm_stream` ("Always runs to completion") continues consuming the LLM to the end. The events accumulate in the unbounded queue and are eventually discarded. The DB save and usage capture still execute correctly — that part works. The cost is LLM tokens for a response no one will read. This is a cost optimization concern, not a correctness issue, but worth noting for high-traffic deployments.

### Recommendation

The DB persistence and headers work in this PR is solid. The remaining two items — **heartbeat injection** and **disconnect-aware streaming** — would complete the fix. They address the production hang directly (keepalive prevents idle timeout) and reduce wasted compute (disconnect polling stops the LLM early).

PR #680 has an implementation of both, along with functional tests that validate the behavior before and after the fix against the real Soliplex app using the `faux` room. Happy to collaborate on merging the remaining pieces into the current architecture.

PR #757, opened 2026-03-17
feat: Feedback room and tools

### LINKED ISSUE #751: AI skill for querying / updating AGUI run feedback1
/cc @runyaga 

Allow a sufficiently-privileged user (in a special room?) to query recent feedback, and mark it as "reviewed" or "resolved" via LLM interaction.

This skill should run entirely server-side, using the `agui.ThreadStorage` facade.

tseaver:
Closes #751.

### DIFF
--- a/src/soliplex/installation.py
+++ b/src/soliplex/installation.py
@@ -39,9 +39,6 @@ NO_AUTH_MODE_USER_TOKEN = {
     "email": "phreddy@example.com",
 }
 
-_THE_AGUI_ENGINE = None
-_THE_AUTHZ_ENGINE = None
-
 
 @dataclasses.dataclass
 class Installation:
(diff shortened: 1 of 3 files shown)


src/soliplex/installation.py
```diff
@@ -38,6 +38,9 @@
     "email": "phreddy@example.com",
 }
 
+_THE_AGUI_ENGINE = None
```
ggozad: Hmm, this looked a bit suspicious. We create a "private" `_THE_AGUI_ENGINE` that is set during the lifespan. 
* The tool module depends on it.
* It makes it necessary to have `No_AGUI_Engine` exception as a check.
* You need to mock it in tests.

Would it be better to put it in `AgentDependencies`? There it would exist for all that need it, for example tools would get it via `ctx.deps.threads_engine`

PR #820, opened 2026-04-02
Automatic thread title generation

ggozad:
Replaces #648 

- Add `titles` module for automatic thread title generation using a configurable LLM agent
- Add `title_agent_config_id` to `InstallationConfig` (opt-in, disabled when not set)
- Hook title generation into `drive_llm_stream`, triggered only after a successful run
- Add title agent config to example installations

### DIFF
--- a/src/soliplex/titles.py
+++ b/src/soliplex/titles.py
@@ -24,28 +24,44 @@ class ThreadTitle(pydantic.BaseModel):
     title: str | None = None
 
 
+def format_message(msg: agui_core.Message) -> str | None:
+    if msg.role not in ("user", "assistant"):
+        return None
+    content = msg.content
+    if content is None:
+        return None
+    if isinstance(content, list):
+        parts = [
+            part.text
+            for part in content
+            if isinstance(part, agui_core.TextInputContent)
+        ]
+        if not parts:
+            return None
+        content = "\n".join(parts)
+    return f"{msg.role}: {content}"
+
+
 def format_messages(messages: list[agui_core.Message]) -> str:
-    lines = []
-    for msg in messages:
-        if msg.role not in ("user", "assistant"):
-            continue
-        content = msg.content
-        if content is None:
-            continue
-        if isinstance(content, list):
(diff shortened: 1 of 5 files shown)


src/soliplex/titles.py
```diff
@@ -0,0 +1,103 @@
...
+Return null if there isn't enough context to generate a meaningful \
+title.\
+"""
+
+
+class ThreadTitle(pydantic.BaseModel):
+    title: str | None = None
+
+
+def format_messages(messages: list[agui_core.Message]) -> str:
```
tseaver: Seems like the flow is a little convoluted?  Maybe hoist the bits for a single message:
```suggestion
def format_message(msg: agui_core.Message) -> str:
    if msg.role in ("user", "assistant"):
        content = msg.content
        if isinstance(content, list):
            parts = [
                part.text
                for part in content
                if isinstance(part, agui_core.TextInputContent)
            ]
            content = "\n".join(parts)
        return f"{msg.role}: {content}") if content else None


def format_messages(messages: list[agui_core.Message]) -> str:
    lines = [
        format_message(msg) for msg in messages
    ]
    return "\n".join(filter(None, lines))
```

src/soliplex/views/agui.py
```diff
@@ -644,6 +648,16 @@ async def drive_llm_stream(
...
                 logfire.info("Stream status: {status}", status=status)
 
+            if status == "FINISHED" and title_agent_config is not None:
+                await titles.maybe_generate_title(
+                    title_agent_config=title_agent_config,
+                    threads_engine=sqla_engine,
+                    room_id=room_id,
+                    thread_id=thread_id,
+                    user_name=user_name,
+                    messages=messages,
```
tseaver: `messages` here will have `assistant`-role messages only from prior runs.  Do we need to be extracting the `assistant`-role messages from the current `event_list` (likely using `agui.parser.EventStreamParser`)?

tseaver (approved):

PR #828, opened 2026-04-06
sandbox + RAG room with per-thread file uploads

ggozad:
Added support for a Docker sandbox room with per-thread file uploads.

New skill config kind `"sandbox"` (`config/skills.py`)  creates a fresh `haiku-skills-sandbox` `Skill` on each access, avoiding shared-state races between concurrent threads. Inherits `_SkillPropertiesFromMetadata` for the full config interface and sets `_factory` on created skills to enable `reconfigure()`.

Per-thread agent caching and sandbox workspace mounting: when `thread_id` is provided, agents are cached per `(room_id, thread_id)` instead of just `room_id`. A new `_reconfigure_sandbox()` helper finds the sandbox skill in the `SkillToolset` registry and calls `reconfigure(workspace=threads_upload_path/thread_id)`, mounting the thread's upload directory at `/workspace` in the Docker container. The `thread_id` is threaded from the AGUI view endpoint through `get_agent_for_room` to agent creation.

Design decisions

- Per-thread uploads: users upload to their thread; each thread gets its own sandbox container with its own workspace mount.
- Fresh `Skill` per request: unlike `EntrypointSkillConfig` which returns a shared instance, `SandboxSkillConfig.skill` creates a new `Skill` each call, making concurrent `reconfigure()` safe.
- Container persistence: the sandbox container lives across runs within a thread via `SandboxState.session_id` in AG-UI state. The mount is set once on first container creation and reused.

Example room: `example/rooms/data-sandbox/` with a room config combining sandbox and RAG skills, file upload tools, and a system prompt for orchestrating between them.

### DIFF
--- a/example/rooms/data-sandbox/prompt.txt
+++ b/example/rooms/data-sandbox/prompt.txt
@@ -1,0 +1,11 @@
+You are a data analysis assistant. You coordinate between a document knowledge base and a Docker sandbox to help users work with their uploaded files.
+
+You have two skills:
+- **rag**: Search and retrieve information from the document knowledge base.
+- **sandbox**: Write and execute Python code in a Docker container. Uploaded files are available at /workspace.
+
+You also have tools to list and read files uploaded to this thread.
+
+Your job is to understand what the user needs, gather relevant context from the knowledge base, and delegate code writing and execution to the sandbox with clear, self-contained requests.
+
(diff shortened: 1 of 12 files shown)


tseaver:
Superseded by the `bwrap_sandbox` skill.

PR #830, opened 2026-04-06
fix: implement server-side OIDC state storage

runyaga:
This PR implements server-side OIDC state and nonce storage in the database, eliminating the dependency on session cookies for the OAuth handshake. This is a robust solution for SameSite cookie issues during cross-site redirects (e.g. CAC/ADFS flows) and is immune to infrastructure-level cookie overrides (like Nginx 'proxy_cookie_flags').

Key changes:
- Added 'oidc_states' table to the database schema.
- Updated OIDC login/callback views to use the database for state/nonce storage.
- Reverted session cookie to 'SameSite=Lax' for improved security posture.
- Pass tokens via URL fragment instead of query parameters to prevent leakage to logs/history.
- Added diagnostic logging for OAuth errors.

### DIFF
--- a/src/soliplex/authz/__init__.py
+++ b/src/soliplex/authz/__init__.py
@@ -78,6 +78,23 @@ class AuthorizationPolicy(abc.ABC):
     ) -> None:
         """Delete the authorization policy for the room"""
 
+    @abc.abstractmethod
+    async def create_oidc_state(
+        self,
+        state: str,
+        nonce: str,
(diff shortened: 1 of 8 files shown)


tseaver (commented):
This design uses the `state` to identify the browser, which is not how the `state` field is expected to be used in OAuth flows.  The state is present in the URL, which means it could be captured and used e.g. for a replay attack.

PR #857, opened 2026-04-10
fix(soliplex): use safe_load in installation

basheer-cloud:
I noticed this while tracing through src/soliplex/config/installation.py. Use safe_load in installation. yaml.load on untrusted input can construct arbitrary Python objects. I kept the patch small and re-ran syntax checks after applying it.

### DIFF
--- a/src/soliplex/config/installation.py
+++ b/src/soliplex/config/installation.py
@@ -93,7 +93,7 @@ def _load_config_yaml(config_path: pathlib.Path) -> dict:
     try:
         with config_path.open() as stream:
             config_yaml = _check_is_dict(
-                yaml.load(stream, yaml.Loader),
+                yaml.safe_load(stream, yaml.Loader),
             )
 
     except Exception as exc:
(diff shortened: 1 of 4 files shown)


tseaver (commented):
Thanks for the patch!  The change which matches the title looks fine.

Adding timeouts to all REST calls seems more debatable:  if we're going to have one, I think should be opt-in via a CLI argument.

src/soliplex/config/installation.py
```diff
@@ -93,7 +93,7 @@ def _load_config_yaml(config_path: pathlib.Path) -> dict:
     try:
         with config_path.open() as stream:
             config_yaml = _check_is_dict(
-                yaml.load(stream, yaml.Loader),
+                yaml.safe_load(stream, yaml.Loader),
```
tseaver, line 96: No second `loader` argument to `safe_load`.
```suggestion
                yaml.safe_load(stream),
```

tseaver (changes requested):
Please run `ruff format` on your branch to resolve the linting issues.

tseaver:
Superseded by #890.

PR #879, opened 2026-04-14
fix: allow use of sqlite databases with existing connection pools

bd-mkt:
support using sqlite database files for testing with connection pools

### DIFF
--- a/tests/unit/test_installation.py
+++ b/tests/unit/test_installation.py
@@ -1,14 +1,14 @@
 import contextlib
 import dataclasses
+import sqlite3
 from unittest import mock
 
 import fastapi
 import pytest
 import sqlalchemy
-import sqlite3
 from ag_ui import core as agui_core
-from sqlalchemy.ext import asyncio as sqla_asyncio
 from sqlalchemy import pool as sqla_pool  # NullPool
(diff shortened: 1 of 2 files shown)


tests/unit/test_installation.py
```diff
@@ -1478,3 +1478,52 @@ async def test_lifespan(
...
+        assert mode == "wal"
+
+    # Engine works and the event listener fires
+    async with engine.begin() as conn:
+        rows = await conn.exec_driver_sql("PRAGMA synchronous")
+        (value,) = rows.fetchone()
+        # NORMAL = 1
+        assert value == 1
+
+    await engine.dispose()
```
tseaver: Likely needs to be handled in a `finally:` section (similarly for the memory version below).  Otherwise a test failure will emit hard-to-track down warnings.  Or else a `contextlib.ExitStack`?

tseaver (changes requested):
Except for the nit I picked on the tests, this change looks great.

tests/unit/test_installation.py
tseaver: 79dacf16

tseaver (approved):

PR #881, opened 2026-04-14
chore: migrate github actions to use uv --frozen

bd-mkt:


tseaver (commented):
x

tseaver (changes requested):
Can we split the changes for the PyPI release out to a separate PR? I'd prefer to land that part just before an actual release, so that we see them run.

tseaver (approved):

PR #882, opened 2026-04-15
Bubble AG-UI events from skill sub-agents

ggozad:


src/soliplex/views/agui.py
```diff
@@ -37,6 +38,36 @@
 depend_the_logger = views.depend_the_logger
 
 
+def _find_skill_toolset(
```
tseaver: This factoring makes me a bit uneasy.  I could wish that we had an `AguiEventStream` class which was agnostic of the toolset, but which exposed its queue such that we could compose it and the toolset queue handling using e.g. a `contextlib.ExitStack`.

ggozad: Relesed hs 0.14.0 with your changes :) 
Adapted this PR to drop the branching.

PR #883, opened 2026-04-15
Support MCP SSE transport

ggozad:
Soliplex mapped kind: "http" to `MCPServerStreamableHTTP`, which sends `POST` requests. MCP servers using the older SSE transport expect `GET` requests, resulting in `405 Method Not Allowed`.

### DIFF
--- a/src/soliplex/mcp_client.py
+++ b/src/soliplex/mcp_client.py
@@ -88,7 +88,44 @@ class HTTP_MCP_Client_Toolset(ai_mcp.MCPServerStreamableHTTP):
         return _filter_tools(offered_tools, self.allowed_tools)
 
 
+class SSE_MCP_Client_Toolset(ai_mcp.MCPServerSSE):
+    def __init__(
+        self,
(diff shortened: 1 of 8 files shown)


src/soliplex/mcp_client.py
```diff
@@ -88,7 +88,44 @@ async def list_tools(self) -> list[mcp_types.Tool]:  # pragma: NO COVER
         return _filter_tools(offered_tools, self.allowed_tools)
 
 
+class SSE_MCP_Client_Toolset(ai_mcp.MCPServerSSE):
```
tseaver: Can we hoist a mixin class which does all this, and make the HTTP and SSE versions derive from it?  The two are identical except for their base class.

tseaver (changes requested):

src/soliplex/mcp_client.py
ggozad: Yep!

tseaver (approved):

PR #892, opened 2026-04-16
fix: avoid 'sqlite3.connect' leaks

### LINKED ISSUE #891: File-based sqlite3 initialization leaks a connection
Two resource warnings show up in tests:  one for the `sqlite3.connect` call inside 'installation._create_async_engine`, and one for the similar call in its tests.  They *don't* appear when running with coverage disabled (grrr...).

tseaver:
Closes #891.


bd-mkt (commented):
looks fine

PR #904, opened 2026-04-20
chore: bump 'haiku.rag-slim >= 0.42.0, < 0.43'

ggozad:
haiku.rag 0.41 renames the RLM skill to the Analysis skill across every layer of its public API: module path (`skills.rlm` -> `skills.analysis`), skill entrypoint (`rag-rlm` -> `rag-analysis`), YAML kind (`haiku.rag.skills.rlm` -> `haiku.rag.skills.analysis`), and state namespace (`rlm` -> `analysis`).

This updates soliplex's `HR_RLM_SkillConfig` to `HR_Analysis_SkillConfig` and propagates the rename through source, tests, docs, example YAMLs, and the generated `schemas/schema.json`.

haiku.rag 0.42 moves the skill lifecycle onto the `haiku.skills>=0.15.0` `lifespan` hook — one `HaikuRAG` client per invocation, state scoped to the current turn, and sandbox variables persisted across `execute_code` calls within a single invocation. The state schema shape is unchanged, so `schemas/schema.json` is identical between 0.41 and 0.42. This PR also bumps `haiku-skills` to `>= 0.15.0, < 0.16`, required by 0.42's lifespan hook.

Also:

- `search.max_context_items` was removed in haiku.rag 0.41; drop from `example/haiku.rag.yaml`.

- `image-generation` is dropped from `scripts/generate_feature_schemas.sh` since it requires the optional `haiku-skills-image-generation` package and previously emitted a null entry when absent; put it back in the jq expression if you want it in the schema.

The RAG skill state shape also changed in 0.41 (`qa_history` / `reports` removed; `citation_index` / `citations` added); the new shape is reflected in the regenerated schema and will flow through to Flutter in a follow-up PR.

### DIFF
--- a/schemas/schema.json
+++ b/schemas/schema.json
@@ -201,10 +201,7 @@
         },
         "citations": {
           "items": {
-            "items": {
-              "type": "string"
-            },
-            "type": "array"
+            "type": "string"
           },
           "title": "Citations",
           "type": "array"
@@ -481,10 +478,7 @@
         },
         "citations": {
           "items": {
-            "items": {
-              "type": "string"
-            },
-            "type": "array"
+            "type": "string"
           },
           "title": "Citations",
           "type": "array"
--- a/schemas/schema.json
+++ b/schemas/schema.json
@@ -2,50 +2,6 @@
   "$schema": "http://json-schema.org/draft-07/schema#",
   "type": "object",
   "properties": {
-    "image-generation": {
-      "$defs": {
-        "GeneratedImage": {
-          "properties": {
-            "prompt": {
-              "title": "Prompt",
-              "type": "string"
-            },
-            "path": {
-              "title": "Path",
-              "type": "string"
-            },
-            "width": {
-              "title": "Width",
-              "type": "integer"
-            },
-            "height": {
-              "title": "Height",
-              "type": "integer"
-            }
-          },
-          "required": [
-            "prompt",
-            "path",
-            "width",
-            "height"
-          ],
-          "title": "GeneratedImage",
-          "type": "object"
-        }
-      },
-      "properties": {
-        "images": {
-          "default": [],
-          "items": {
-            "$ref": "#/$defs/GeneratedImage"
-          },
-          "title": "Images",
-          "type": "array"
-        }
-      },
-      "title": "ImageState",
-      "type": "object"
-    },
     "rag": {
       "$defs": {
         "Citation": {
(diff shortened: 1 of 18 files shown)


src/soliplex/config/skills.py
```diff
@@ -461,7 +461,7 @@ def skill(self) -> hs_models.Skill:
         FilesystemSkillConfig,
         EntrypointSkillConfig,
         HR_RAG_SkillConfig,
-        HR_RLM_SkillConfig,
+        HR_Analysis_SkillConfig,
         BwrapSandboxSkillConfig,
```
tseaver, line 465: Do we need to add a forward-compat alias for the `"rag"` kind?  So that Chad doesn't have to do a batch update before testing?

schemas/schema.json
```diff
@@ -2,50 +2,6 @@
   "$schema": "http://json-schema.org/draft-07/schema#",
   "type": "object",
   "properties": {
-    "image-generation": {
```
tseaver, line 5: I wilsh we weren't carrying this file at all, honestly.

src/soliplex/config/skills.py
ggozad: You mean migrating the DBs? This should be done already, 0.41.0 does not need migration, just update.
Or maybe you are referring to something else?

schemas/schema.json
ggozad: Oh yes, it won't be missed. I think this is a remnant of that json generator for dart? Do we actually still need it?

src/soliplex/config/skills.py
tseaver: I was thinking about the "hundreds" of room configs which use the `rlm` kind. :)

In fact, looking at our customer configs, I don't see *any* using the `rlm` skill kind.  The only one I find is in the `example/analysis` room in this repo.

schemas/schema.json
tseaver: I don't know,

I'm pretty sure the Flutter front-end code will need some changes, though.  I see you created https://github.com/soliplex/frontend/pull/133, and `git grep "schema\.json"` in that repo doesnt return results.

src/soliplex/config/skills.py
ggozad: Yeah the RLM was pretty experimental and not really usable so I had also advised not to use it, except to play with it.

PR #914, opened 2026-04-21
Changes for making it easy to onboard

frapell:
# Onboarding Soliplex

## Backend

Main repo to clone, [this one](https://github.com/soliplex/soliplex):
```
$ git clone git@github.com:soliplex/soliplex.git
```

Once cloned, copy the `.env.example` as `.env` and assign your values to the relevant environment variables.
Next, build:
```
$ cd soliplex/
$ cp .env.example .env
$ docker compose build
```

Once built, start the process:
```
$ docker compose up
```

### Indexing documents automatically

Automatically, a `haiku.rag.lancedb` will be created in `db/rag` folder. And automatically, it will index `PDF` documents from inside the `documents` folder. New files will be created next to each `PDF` file with a `.index` extension will be created so the script knows this `PDF` was already indexed. If you want to have the automatic process re-index this `PDF`, simply delete the `.index` file and restart.

### Running haiku.rag commands manually

You can jump inside the docker container with `docker compose run soliplex_backend /bin/bash` and manually run `haiku-rag` commands.

## Frontend

In order to add the Flutter frontend to the mix, you should clone the [Flutter frontend](https://github.com/soliplex/frontend) into `src/`:
```
$ cd src/
$ git clone git@github.com:soliplex/frontend.git
```

Once cloned, go back to the main `soliplex` repo, uncomment the relevant lines in `docker-compose.yaml`:

```
  soliplex_frontend:
    build:
      context: ./src/frontend
      dockerfile: Dockerfile
      target: dev
    ports:
      - "9000:9000"
    volumes:
      - ./src/frontend:/app
    networks:
      - soliplex_net
```

And build again
```
$ docker compose build
```

Once finished, start all
```
$ docker compose up
```

Once up, you can visit http://localhost:9000/ in your browser. The soliplex backend can be reached at http://localhost:8000

## Chatbot widget

In order to add the chatbot widget to the mix, you should clone the [chatbot repo](https://github.com/soliplex/chatbot) into `src/`:
```
$ cd src/
$ git clone git@github.com:soliplex/chatbot.git
```

Once cloned, go back to the main `soliplex` repo, uncomment the relevant lines in `docker-compose.yaml`:

```
  chatbot_builder:
    build:
      context: ./src/chatbot
      dockerfile: Dockerfile
    ports:
      - "35729:35729"
    volumes:
      - ./src/chatbot:/app
      - /app/node_modules
    environment:
      - ESBUILD_POLL=true
    networks:
      - soliplex_net

  chatbot_dev:
    build:
      context: ./src/chatbot
      dockerfile: Dockerfile
    command: npm run dev
    ports:
      - "3000:3000"
    volumes:
      - ./src/chatbot:/app
      - /app/node_modules
    environment:
      - WATCHPACK_POLLING=true
    networks:
      - soliplex_net

  chatbot_widget:
    image: nginx:alpine
    ports:
      - "8080:80"
    volumes:
      - ./src/chatbot/public:/usr/share/nginx/html:ro
      - ./src/chatbot/nginx-widget.conf:/etc/nginx/conf.d/default.conf:ro
    depends_on:
      - chatbot_builder
    networks:
      - soliplex_net
```

And build again
```
$ docker compose build
```

Once finished, start all
```
$ docker compose up
```

At this point, you can access the widget in "dev mode" at http://localhost:3000/ where changes to the code are applied live, as a normal React App.
In addition, you can access http://localhost:8080/ to use it through Nginx, with the compiled JS embedded. Changes to the code, will reload the whole page.


docker-compose.yaml
```diff
@@ -1,65 +1,85 @@
...
     build:
       context: .
       dockerfile: Dockerfile
-    #network_mode: "host" # only works on linux
-    ports:
-      - "8000:8000"
-    extra_hosts:
-      - "host.docker.internal:host-gateway"
-    # Needed to run 'bubblewrap' sandbox
-    privileged: true
```
tseaver, line 13: These two lines are important:  please restore them.

docker-compose.yaml
```diff
@@ -1,65 +1,85 @@
...
-
-  # Example: development service with hot reloading and bind-mounted source.
-  #
-  #   docker compose up soliplex_dev
-  #
-  # On Linux, set APP_UID/APP_GID in .env or your shell so that
-  # container-written files match your host user:
-  #
-  #   APP_UID=$(id -u) APP_GID=$(id -g) docker compose up soliplex_dev
-  #
```
tseaver, line 36: We don't want to be running in the container as root:  it litters root-owned files (e.g. file uploads, sandbox scripts) into the bind-mount host system directories.

docker-compose.yaml
```diff
@@ -1,65 +1,85 @@
...
-    privileged: true
     env_file:
       - .env
-    command:
-      - soliplex-cli
-      - serve
-      - --host=0.0.0.0
-      - --reload=both
-      - --no-auth-mode
-      - /app/installation/minimal.yaml
```
tseaver, line 58: Do *not* expect the `Dockerfile` command to be usable:  the `docker compose` environment is where the user configures this stuff, including deciding which installation config to be used, whether to use auth mode, etc.

docker-compose.yaml
```diff
@@ -1,65 +1,85 @@
...
+  #   build:
+  #     context: ./src/frontend
+  #     dockerfile: Dockerfile
+  #     target: dev
+  #   ports:
+  #     - "9000:9000"
+  #   volumes:
+  #     - ./src/frontend:/app
+  #   networks:
+  #     - soliplex_net
```
tseaver, line 84: When developing in this repo, we could actually make use of a published image for the frontend.  Folks who want to hack on the frontend should to that within the context of that repo, not here.

docker-compose.yaml
```diff
@@ -1,65 +1,85 @@
...
+
+  # If you want to test the embeddable chat widget (https://github.com/soliplex/chatbot/)
```
tseaver, line 86: Likewise here:  don't expect folks to do development on the chatbot in this repo.

tseaver (changes requested):
This review is partial:  I haven't looked yet at the `entrypoint.sh` bits.

docker-compose.yaml
frapell: Doing this is a huge security hole... I added it back, but I think we should find a way around it

docker-compose.yaml
frapell: I disagree with this... I prefer to have the image working properly OOTB, and then have the option to add my own command if I need to. If you insist, I'll add it back, but I would rather not

docker-compose.yaml
frapell: Having this commented, is a huge win for someone simply trying to use it, I would prefer to keep it as a convenience

docker-compose.yaml
frapell: Having this commented, is a huge win for someone simply trying to use it, I would prefer to keep it as a convenience

frapell:
@tseaver Applied some of your suggestions, and I force pushed becausse I rebased against main (Mainly for the newer versions for `bubble-sandbox`). @runyaga I added an example (`server_time`) to include your own python module that should be executed in a sandbox. Also, updated the docs, which I will add in the next comment

frapell:
# Onboarding Soliplex

## Backend

Main repo to clone, [this one](https://github.com/soliplex/soliplex):
```
$ git clone git@github.com:soliplex/soliplex.git
```

Once cloned, copy the `.env.example` as `.env` and assign your values to the relevant environment variables.
Next, build:
```
$ cd soliplex/
$ cp .env.example .env
$ docker compose build
```

Generate your `installation` directory. Either start from scratch, or copy the `example` one:
```
cp -R example installation
```
Notice that by default, `soliplex` will look for a `installation.yaml` file inside the `installation` folder. You can either change this with your own command for starting `soliplex-cli` or simply by making sure your config is properly made inside this file.

Once built, start the process:
```
$ docker compose up
```

### Debugging

`debugpy` is included as a `dev` dependency for soliplex, and it should listen on port `5678`. There is already a `.vscode/launch.json` included at the root of the repo, so if you are using VSCode, after the process is up, you can go to the `Debug` tab in VSCode and connect to the running process.

### Sandboxed python modules

You can include Python modules, to be executed in rooms in their own sandbox. A "server_time" is included as example at `sandbox/environments/server_time`, configured to be used in the `example/rooms/server_time` room. It is important to notice that a proper description should be included in the `pyproject.toml` for your module, in order for the AI agent to pass the correct `environment_name` to the `execute_script` call.
Notice that changes to these modules will require a `docker compose build` in order for the code to be updated and used.

### Indexing documents automatically

Automatically, a `haiku.rag.lancedb` will be created in `db/rag` folder. And automatically, it will index `PDF` documents from inside the `documents` folder. New files will be created next to each `PDF` file with a `.index` extension will be created so the script knows this `PDF` was already indexed. If you want to have the automatic process re-index this `PDF`, simply delete the `.index` file and restart.

### Running haiku.rag commands manually

You can jump inside the docker container with `docker compose run soliplex_backend /bin/bash` and manually run `haiku-rag` commands.


## Frontend

In order to add the Flutter frontend to the mix, you should clone the [Flutter frontend](https://github.com/soliplex/frontend) into `src/`:
```
$ cd src/
$ git clone git@github.com:soliplex/frontend.git
```

Once cloned, go back to the main `soliplex` repo, uncomment the relevant lines in `docker-compose.yaml`:

```
  soliplex_frontend:
    build:
      context: ./src/frontend
      dockerfile: Dockerfile
      target: dev
    ports:
      - "9000:9000"
    volumes:
      - ./src/frontend:/app
    networks:
      - soliplex_net
```

And build again
```
$ docker compose build
```

Once finished, start all
```
$ docker compose up
```

Once up, you can visit http://localhost:9000/ in your browser. The soliplex backend can be reached at http://localhost:8000


## Chatbot widget

In order to add the chatbot widget to the mix, you should clone the [chatbot repo](https://github.com/soliplex/chatbot) into `src/`:
```
$ cd src/
$ git clone git@github.com:soliplex/chatbot.git
```

Once cloned, go back to the main `soliplex` repo, uncomment the relevant lines in `docker-compose.yaml`:

```
  chatbot_builder:
    build:
      context: ./src/chatbot
      dockerfile: Dockerfile
    ports:
      - "35729:35729"
    volumes:
      - ./src/chatbot:/app
      - /app/node_modules
    environment:
      - ESBUILD_POLL=true
    networks:
      - soliplex_net

  chatbot_dev:
    build:
      context: ./src/chatbot
      dockerfile: Dockerfile
    command: npm run dev
    ports:
      - "3000:3000"
    volumes:
      - ./src/chatbot:/app
      - /app/node_modules
    environment:
      - WATCHPACK_POLLING=true
    networks:
      - soliplex_net

  chatbot_widget:
    image: nginx:alpine
    ports:
      - "8080:80"
    volumes:
      - ./src/chatbot/public:/usr/share/nginx/html:ro
      - ./src/chatbot/nginx-widget.conf:/etc/nginx/conf.d/default.conf:ro
    depends_on:
      - chatbot_builder
    networks:
      - soliplex_net
```

And build again
```
$ docker compose build
```

Once finished, start all
```
$ docker compose up
```

At this point, you can access the widget in "dev mode" at http://localhost:3000/ where changes to the code are applied live, as a normal React App.
In addition, you can access http://localhost:8080/ to use it through Nginx, with the compiled JS embedded. Changes to the code, will reload the whole page.

ggozad:
The bash entrypoint reinvents what is already in `haiku.rag-slim` and introdcues a few  bugs in the current shape.

1. It creates `db/rag/haiku.rag.lancedb`, but `example/rooms/bwrap_sandbox/room_config.yaml` reads `rag_lancedb_stem: "rag"`, which resolves to `db/rag/rag.lancedb`. The room never sees what the entrypoint indexed.
2. It calls `haiku-rag init` and `add-src` without `--config`, so it uses haiku-rag's built-in defaults instead of `example/haiku.rag.yaml`. The embedding model and vector dim will not match what soliplex queries with.
3. `add-src` failures are wrapped in `if ... then touch indexed; else echo skipping`, so PDF parsing failures are silent. `haiku.rag-slim` ships `docling-core` only, not `docling`, so PDFs need `docling-serve` running to be ingested at all, and it is not in compose. Combined with the silent failure path, the user sees a clean startup but an empty RAG.


The replacement is two compose services and a config knob.

```yaml
docling-serve:
  image: quay.io/docling-project/docling-serve:latest
  ports:
    - "127.0.0.1:5001:5001"
  environment:
    - DOCLING_SERVE_ENABLE_UI=1
  healthcheck:
    test: ["CMD", "curl", "-f", "http://localhost:5001/health"]
    interval: 30s
    timeout: 10s
    retries: 3
    start_period: 40s
  networks:
    - soliplex_net

soliplex_indexer:
  build:
    context: .
    dockerfile: Dockerfile
  command:
    - haiku-rag
    - serve
    - --monitor
    - --config=/app/installation/haiku.rag.yaml
    - --db=/app/db/rag/rag.lancedb
  env_file:
    - .env
  depends_on:
    docling-serve:
      condition: service_healthy
  volumes:
    - ./installation:/app/installation:ro
    - ./db:/app/db
    - ./documents:/app/documents
  networks:
    - soliplex_net
```

`soliplex_backend` should also pick up `depends_on: { docling-serve: { condition: service_healthy } }`.

In `example/haiku.rag.yaml`, add:

```yaml
providers:
  docling_serve:
    base_url: http://docling-serve:5001
```

Plus a `monitor:` block pointing at `/app/documents` with sensible include patterns.

The existing comment in `example/haiku.rag.yaml` that references https://github.com/ggozad/haiku.rag/issues/132 is stale. The issue was fixed in haiku.rag 0.15.0 and we pin `>= 0.43.0`. Please delete that comment.

This way we get continuous file watching instead of just on container start, modify and delete events, dedup via `updated_at`

ggozad:
The goal of making soliplex easy to pick up is not an easy one but important, and a few of the changes here do that well. There is a tension in the PR that I think is worth talking through before merge.

A new user's first ten minutes after git clone should have as few steps and as few moving parts as possible. Some of what is in this PR pushes the other way like `debugpy` for instance. I.e. I think this should aim for people who want to try this out, not developers who will customize heavily, develop or contribute.
git clone, cp .env.example .env, docker compose up should be the whole story.

Having said that, these are the parts that I think should be addressed regardless:

- Removing the `production` Dockerfile stage and the non-root `soliplex` user. These are deployment-posture choices, not onboarding. Keeping production as the default `docker build` target is intentional, and dropping non-root is painful to add back later.
- Switching from `uv sync --frozen` to `pip install -e .`. This loses lockfile pinning and reproducible builds.
- Removing `HEALTHCHECK` from both stages.
- Removing `--reload=both` from the dev `CMD`. The dev target's whole job is hot reload.
- Mounting `/var/run/docker.sock` into the backend container. That is effectively root on the host. Needs explicit justification or drop.
- `pip install --user` in a multi-stage build is unusual. Prefer a venv if we change the install path at all.
- `ONBOARD.md` says `docker compose run soliplex_backend /bin/bash` to enter "the container". `run` creates a new one. For the running one, use `exec`.
- `ONBOARD.md` duplicates much of `docs/docker.md` from the mkdocs site. Two sources of truth will drift.

frapell:
@ggozad Thanks a lot for your feedback, I just pushed the changes and here are my responses to your points:

- `debugpy` I agree with your statement, and in fact, this is simply a conveninence to have. It is available if someone wants to use it, so they don't need to manually add the dependency and install, and doesn't affect the execution if someone is not using it. That said, there seems to be an issue when you have hot reload enabled, and since you prefer that, I removed `debugpy` from being automatically initialized, but left it as a `dev` dependency anyway, just in case.
- As per the `Dockerfile` changes, I basically restored the version that was in `main`, so that means back to `uv sync --frozen` instead of `pip`
- I also restored the `HEALTHCHECK` line. I personally prefer to have this in the docker-compose file, but I believe there is no real difference in having it here, as it can be disabled if wanted to anyway.
- The docker socket was being included as the intention to run sandboxed code was intended to do with `docker-in-docker`, but my understanding is that now this is done using `bubblewrap` (Which introduces the `privileged: true` which does provide root access to the host ;) ) so it is no longer needed.
- The `ONBOARD.md` was intended to avoid updating the comment here with the intended usage. I guess we can remove it once this branch functionality is finished, and only keep the documentation in `docs/docker.md`

Bottom line, this branch does work as you suggested, i.e. 

1. Clone repo.
2. Initial configuration (`installation/` directory, envs in `.env` and init rag db)
3. compose up.

ggozad:
Thanks for the rework, this is a big improvement. I took notes so I apologise for not doing individual comments on file :( ... 

Some small things I would fix:

* `cp -R example installation`

Your bottom line lists step 2 as "initial configuration (installation/ directory, envs in .env, init rag db)". Today on `main`, `git clone`, `cp .env.example .env`, `docker compose up` is the whole story, because the compose file mounts `./example:/app/installation`. This PR moves the mount to `./installation:/app/installation` and asks the user to copy the example directory first.

I would prefer the default mount stay on `./example`. Anyone doing real customization is already going to override the mount in their own compose file or change the bind path. A first-time visitor should not have to copy the example directory just to see soliplex run. If we keep the new layout, we trade a working `docker compose up` for a `cp -R` step on every fresh checkout, and that is exactly the friction this PR is meant to remove.

* Leftovers from the rework

- `entrypoint.sh` is dead code now. The Dockerfile revert dropped its `ENTRYPOINT` directive, so nothing runs it. Please delete the file.
- The "Indexing documents automatically" section in `ONBOARD.md` still describes the old mechanism (`.index` markers, the `haiku.rag.lancedb` path). The new flow is "drop files in `documents/`, the indexer watches them via `serve --monitor`". This section needs a rewrite to match what the PR now does.
- The one-off `docker compose run --rm soliplex_indexer haiku-rag ... init ...` step in `ONBOARD.md` could be folded into the indexer service so init happens idempotently on first start. Then the user just runs `docker compose up` and the indexer takes care of itself.

* Nitpicks

- `TZ=Pacific/Auckland` is still hardcoded in `docker-compose.yaml` under `soliplex_backend.environment` with the `XXX:` comment. Please remove.
- `.vscode/launch.json` is editor-specific config. I would either drop it, or if it is genuinely useful as documentation move it under `docs/` as a snippet.
- `5678:5678` should bind to `127.0.0.1:5678:5678` so the debug port is not exposed beyond the host.
- `pyproject.toml` adds `"debugpy"` with no version pin, inconsistent with its neighbors. If we keep it, please pin it.
- `sandbox/environments/server_time/src/server_time/__init__.py` docstring says `"""Hour ..."""`. Should be `"""Server time ..."""`.
- `example/rooms/server_time/room_config.yaml` uses `default_environment_name`. Could you confirm this key is supported by the `bubble-sandbox` version we pin? I have not checked.

docker-compose.yaml
tseaver: It is required for any container which use `bwrap` for sandboxing: in production, we may move to a separate container running an updated version of @bd-mkt 's `exec_wrap`:  if so, *that* container would then need the flag.

frapell:
@ggozad Thanks again a lot for all this :) latest changes:

- Removed the `ONBOARD.md` and directly changed the `docs/docker.md` so there is only 1 source of truth. This file was updated with all the changes on this branch.
- Replaced the mount to be `./example`. The idea to keep it separate was to avoid accidental conflict with changes to the folder, or accidental commits with changes to it.
- Removed the `entrypoint.sh` and `.vscode/launch.json` as well as the `debugpy` dep.
- Commented the `TZ=Pacific/Auckland` env, but left it in the `docker-compose.yaml` as it is a quick way to test the example sandboxed python module.
- Removed the `default_environment_name` as I included it when was trying to figure out how the `bubblewrap` worked, thinking this was a way to "force" which environment to use, but in reality the agent first checks which environment is appropriate to pass, and then calls `execute_script` with it, and there is no apparent way to force this.

tseaver:
Superseded by `soliplex-template v0.16`, which allows generating a full-service `compose` stack with `soliplex` cloned and running in dev mode.  See: https://soliplex.github.io/soliplex-template/getting-started/tutorials/backend-dev-mode/

PR #921, opened 2026-04-21
chore: Update bwrap skill description to better instruct on order of operation and functionality

Cwiesen:


src/soliplex/skills/bwrap_sandbox/SKILL.md
```diff
@@ -1,73 +1,100 @@
...
-  shell commands unless you specifically need shell functionality.
-- Write self-contained scripts that print their output.
-- Output files (CSVs, plots) written to `/sandbox/work/` persist
-  across calls within the same run.
-- If a script fails, read the error, fix the code, and retry.
+1. **Pick an environment.** If the task needs third-party packages,
+   call `list_environments` and choose the one whose description
+   matches the task. Pass its `name` as `environment_name` to
+   `execute_script`. If the task is pure stdlib, you can skip this
+   and use the default.
```
tseaver, line 62:
```suggestion
1. **Pick an environment.** When first using the sandbox,
   call `list_environments` to get information on the available
   environments,    If the task needs third-party packages,
   choose the one whose description and dependencies match
   task. Pass its `name` as `environment_name` to
   `execute_script`.
```

src/soliplex/skills/bwrap_sandbox/SKILL.md
```diff
@@ -1,73 +1,100 @@
...
+       p = pathlib.Path(vol)
+       if p.exists():
+           for f in sorted(p.rglob('*')):
+               if f.is_file():
+                   print(f, f.stat().st_size)
+   ```
+
+   Then read specific files you actually need. Always check the
+   `room` volume too — it may contain rules or formulas required for
+   a correct answer, not just the obvious `thread` inputs.
```
tseaver, line 80:
```suggestion
   Then read specific files you actually need. Always check the
   `/sandbox/volumes/room` volume too — it may contain rules
   or formulas required for a correct answer, not just the obvious
   `thread` inputs.
```

Issue #935, opened 2026-04-24
Quirks identified while documenting CLI commands


tseaver:
# `soliplex-cli` Quirks and Unusual Behavior

Non-obvious or surprising behaviors of the `soliplex-cli` subcommands,
collected during a documentation pass over `src/soliplex/cli.py`. These
are not bugs (most of them) — just things an operator would likely
misjudge from the command names, help text, or natural-language
reading of the docs.

## Semantic footguns — the command does less (or other) than it sounds like

- **`clear-room-authz` without `--make-room-private` *opens* the room.**
  Deleting the `RoomPolicy` row reverts the room to its default state,
  which is public-to-all-authenticated-users. This is the inverse of
  most admin operators' assumption. (`cli.py:1160-1169`,
  `authz/persistence.py:122-127`.)
- **First `add-room-user` call against a previously public room flips
  it to private.** Creating an ACL entry necessarily creates a policy
  row, which carries `default_allow_deny=DENY`. Everyone who previously
  had access loses it in one step. (`cli.py:1214-1234`,
  `authz/schema.py:93-95`.)
- **`pull-models --ollama-url URL` filters, it does not inject.**
  Passing a URL the installation doesn't reference yields an empty
  model set, not an on-demand pull. (`cli.py:713-716`.)
- **`serve`'s `-h` is `--host`, not `--help`.** The `-h`/`--help` alias
  is only bound at the top-level `soliplex-cli` callback; the `serve`
  subcommand overrides `-h` for host. (`cli.py:51` vs `cli.py:167`.)
- **`add-admin-user` has no dedup.** Inserts unconditionally; two
  invocations with the same email produce whatever the DB's uniqueness
  constraint dictates (success + duplicate row, or failure — depending
  on schema). (`cli.py:1061-1064`.)
- **`add-room-user` only knows one shape.** It creates ALLOW-by-email
  entries only — no way to create DENY, `everyone`, `authenticated`,
  or `preferred_username` entries from the CLI. (`cli.py:1228-1232`.)

## Exit-status traps — "success" when things failed or were no-ops

- **Every `list-*` command exits `0` regardless of findings.** Missing
  secrets, missing env vars, broken skills, RAG DBs that won't open,
  or a totally empty configuration — the listing commands report
  inline and return 0. Only `check-config` exits non-zero on problems.
- **`pull-models` exits `0` even when some pulls failed.** Per-model
  failures update the summary's success/total ratio but do not
[...]

PR #944, opened 2026-04-28
refactor: CLI `audit` subcommands

### LINKED ISSUE #936: CLI: group related subcommands
As with #935, noted while documenting the CLI commands.  Group them as follows for clarity, and to reduce the top-level help list.  For all of these, leave hidden subcommands behind with the old names (and backward compatible arguments, if they change).
- [ ] Make the `list-*` commands sub-commands of a new `audit` group, e.g. `soliplex-cli audit rooms`, etc.; improve consistency (`-q` suppresses unnecessary output, errors return non-zero, etc.) .  `check-config` could become `audit all`, perhaps?
- [ ] Make the `*admin-user*` commands subcommands of a new `admin-users` group.
- [ ] Make the room authz commands (including `add-room-user` subcommands of a new `room-authz` group.

tseaver:
Closes #936.

I'm likely to tackle some of the issues raised in #935 on this branch as well:

- [x] All the `audit` subcommands will grow a `-q` mode (like `audit installation`, FKA `check-configuration`, added in PR #925).  In this mode, they will not emit any extraneous output:  instead, the will exit with `0` status if no errors are found, and exit with `1` status if errors are found, emitting JSON on stdout describing the errors.
- [x] I will harmonize the checking done in `audit installation` vs. `audit {secrets,environment,....skills}`:  they should be detecting and reporting the same errors.
Deferred to another day:
~~I'll normalize the CRUD bits in the `admin-users` group, and address quirks from that issue.~~
~~I may need to rethink the `room-users` group a bit.``


tseaver:
Note to reviewers:  this PR is effectively unreviewable at the level of the Python code, for which I apologize.  Even commit-by-commit is likely too painful for words.

Instead, please look at the docs updates, in particular the [CLI docs changes].  It would likely be useful to check out the `cli-reorg` branch, run `uv sync --all-groups` there, and then start poking at `soliplex-cli --help` and related to get a feel for it. 

By far the biggest chunk is consolidating `config-changes` and the `list-*` commands under an `audit` command group, applying a ton of normatization:

- The `--quiet` flag at the group level disables all "chatty" / human-focused output:  each command returns either no output with status code `0` if no errors were found, or a JSON mapping of the errors and status code `1`.
- The `audit` command group is an alias for `audit all`, which is an improved superset of the earlier `check-config` command (still present as a hidden alias for `audit all`).
- The `audit secrets` subcommand replaces `list-secrets` (and similarly for the other `list-*` commands: the older forms also remain as hidden aliases.
- The `serve` command is basically unchanged, except that the short-form alias for `--host` is now `-H` (because `-h` is for help).  AFAICT we haven't use the `-h` form anywhere in configs / scripts.

I did the refactoring slice-by-slice with Claude, with docs updates as we went, so I'm pretty confident the CLI docs are comprehensive.

PR #983, opened 2026-05-18
S3 support

bd-mkt:
Adds support for s3 storage for haiku rag databases. Supports configuration of access parameters via boto3.


tseaver:
See #1246.

PR #984, opened 2026-05-19
feat: 'soliplex-cli shell <installation_path>'

tseaver:
Loads 'the_installation' from the supplied path and drops into a Python REPL prompt for debugging.

Reviewers:  I wrote this to ease querying a configuration as loaded.  It *does* include the ability to see resolved secrets, but given that the user has shell access on the system, I don't think that is a new exposure.  Maybe if we were supporting some kind of third party secret storage, it would be problematic.

### DIFF
--- a/src/soliplex/cli/main.py
+++ b/src/soliplex/cli/main.py
@@ -1,7 +1,9 @@
 from __future__ import annotations
 
+import code
 import json
 import pathlib
+import sys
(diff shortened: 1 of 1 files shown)


tseaver:
Given that this PR is a convenience, rather than a new capability (an actor with shell access can already to the same steps in a bare Python REPL), I think we're good to merge.

Issue #989, opened 2026-05-20
Migrate docs to `zensical`


tseaver:
See https://github.com/ggozad/haiku.rag/pull/380.

tseaver:
Via #998.

PR #1000, opened 2026-05-20
chore: replace mkdocs.yml with zensical.toml

mcdonc:
## Summary
- Replace `mkdocs.yml` with native `zensical.toml` config
- Follows up on PR #998 which swapped the deps

## Test plan
- [x] `zensical build` — clean build, no issues, 2.5s

See also https://github.com/soliplex/soliplex/issues/989


tseaver (approved):
I can't see a win for TOML over YAML, myself, but if this change makes maintaining Zensical docs easier in the future, I'm OK with it.

PR #1031, opened 2026-05-30
feat: add 'skill_versions upgrade'

tseaver:
... upgrades installed skill in place to a given tag
   ('latest' by default).


tseaver:
soliplex/soliplex-skills#25 tracks the Windows furball.

PR #1076, opened 2026-06-16
feat: per-thread last activity for unread markers

WilliamKarolDiCioccio:
Follow-up to #1074. That PR added the room-level activity stats (`GET /v1/stats/rooms`, `RoomStats.last_activity`, `get_rooms_last_activity`); this extends the same surface down to the **thread** level so clients can mark individual threads with unseen activity, not just rooms.

## What

- **`AGUI_Thread.last_activity`** — most recent run activity in the thread (finish, or start while unfinished), or `null` when it has no runs. Added to the thread-listing response (`GET /v1/rooms/{room_id}/agui`), filled by one grouped query — no replaying of each threadʼs runs. Distinct from `created` (the threadʼs birth).
- **`ThreadStorage.get_threads_last_activity`** — grouped `max(coalesce(Run.finished, Run.created))` by thread within a room, scoped to the user. Mirrors `get_rooms_last_activity`, reusing the shared `_LAST_ACTIVITY` expression and `_as_utc` normalization.

## Tests

100% branch coverage on the new code; new persistence tests (empty + grouping/scoping) and a thread-listing view assertion. The pre-existing `test_lifespan` failures are an environment issue (they fail identically on a clean `main`), unrelated to this change.

## Client

The Soliplex frontend already reads `last_activity` off both the batch room stats and the thread listing, so this lights up thread-level unread dots with no client change.

### DIFF
--- a/src/soliplex/models.py
+++ b/src/soliplex/models.py
@@ -825,12 +825,18 @@ class AGUI_Thread(pydantic.BaseModel):
     created: datetime.datetime | None = KW_ONLY_NONE
     metadata: AGUI_ThreadMetadata | None = KW_ONLY_NONE
 
+    # Latest run activity in the thread (finish, or start while unfinished),
+    # or None when it has no runs. Distinct from 'created' (the thread's
+    # birth); lets clients mark threads with unseen activity.
+    last_activity: pydantic.AwareDatetime | None = KW_ONLY_NONE
+
     @classmethod
     def from_thread(
         cls,
         a_thread: agui_package.Thread,
         a_thread_meta: AGUI_ThreadMetadata,
         a_thread_runs: AGUI_Runs = None,
+        a_thread_last_activity: datetime.datetime | None = None,
     ):
         return cls(
             room_id=a_thread.room_id,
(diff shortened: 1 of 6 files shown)


src/soliplex/models.py
```diff
@@ -825,19 +825,26 @@ class AGUI_Thread(pydantic.BaseModel):
     created: datetime.datetime | None = KW_ONLY_NONE
     metadata: AGUI_ThreadMetadata | None = KW_ONLY_NONE
 
+    # Latest run activity in the thread (finish, or start while unfinished),
+    # or None when it has no runs. Distinct from 'created' (the thread's
+    # birth); lets clients mark threads with unseen activity.
+    last_activity: pydantic.AwareDatetime | None = KW_ONLY_NONE
```
tseaver, line 831: While it is technically possible to create a `Thread` record in the database without any child `Run` records, there is no path in the application which allows it:  *every* therad gets its first `Run` added in the `POST /api/vi/rooms/{room_id}/agui` [endpoint view](), which calls [`the_threads.new_thread`](https://github.com/soliplex/soliplex/blob/8dfd4480dc1bde1dfc209ef3c38eab49e721d775/src/soliplex/agui/persistence.py#L189-L191):
```python
            async with session.begin_nested():
                run = agui_schema.Run(thread=thread)
                session.add(run)
```

So the net effect is that you will get at least the `created` date for the initial run here, never `None`.

Issue #1080, opened 2026-06-20
CLI: route `admin-users` subcommands through `AuthorizationPolicy`


tseaver:
Refactor `soliplex.cli.admin_users` so each subcommand performs its database work through `soliplex.authz.persistence.AuthorizationPolicy` (async), driven by `asyncio.run()`, instead of the current direct synchronous-session writes.

### `AuthorizationPolicy` additions — `src/soliplex/authz/persistence.py`

The CLI operates on arbitrary `json_path` discriminators and supports bulk replace, which the current email-only methods don't cover. Add:

- `add_admin_user_by_json_path(json_path: str)` — insert; raise `AdminUserExists` if present.
- `remove_admin_user_by_json_path(json_path: str)` — delete; raise `NoSuchAdminUser` if absent. Must also match a stored entry whose `json_path` no longer compiles (supports `delete --allow-invalid-json-path`).
- `clear_admin_users()` — delete all rows.
- A raw listing (extend `list_admin_users()` or add `list_admin_user_json_paths()`) returning the stored `json_path` strings unmodified, so the CLI renders them via its existing `_admin_display`.

Reduce the existing email-keyed `add_admin_user` / `remove_admin_user` to thin wrappers that build the email `json_path` and delegate.

Update the CLI accordingly.

PR #1083, opened 2026-06-26
feat: 'authz.AuthorizationPolicy' parity w/ `admin-users` CLI

tseaver:
Prefer use of JSONPath discriminators over email addresses.

Existing method taking 'email' argments are now deprecated shims.

Toward #1080.

### DIFF
--- a/src/soliplex/models.py
+++ b/src/soliplex/models.py
@@ -898,7 +898,7 @@ RoomPolicyMap = dict[str, RoomPolicy | None]
 
 
 class InstallationAuthorization(pydantic.BaseModel):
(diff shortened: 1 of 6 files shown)


src/soliplex/models.py
```diff
@@ -898,7 +898,7 @@ class RoomPolicy(pydantic.BaseModel):
 
 
 class InstallationAuthorization(pydantic.BaseModel):
-    admin_user_emails: list[str] = pydantic.Field(default_factory=list)
+    admin_user_discriminators: list[str] = pydantic.Field(default_factory=list)
```
tseaver, line 901: Note: the frontend does not use the `GET /v1/installation/authz` endpoint, which is the only use of this model.

PR #1105, opened 2026-06-29
feat(agui): expose per-event timestamps for live and replay

91jaeminjo:
## What & why

Backend half of the frontend chat-timestamps feature: expose each AG-UI event's per-event creation time to clients on the wire, for **both live streaming and history replay**, so a message's real time is available per-event instead of only at run granularity.

The AG-UI protocol already defines `BaseEvent.timestamp: Optional[int]` (ms since epoch) and clients already parse it, so **no protocol or generated-schema change is needed** — we just populate it.

## Changes

1. **`agui/util.py`** — new `_epoch_ms(datetime) -> int` helper (datetime → int ms; naive datetimes assumed UTC, since SQLite returns zoneless datetimes on read).
2. **`agui/schema.py`** — `RunEvent.to_agui_model()` fills `event.timestamp` from the DB column `created` (via `_epoch_ms`), **fill-only** (only when the event carries no timestamp and `created` is set). This is the single chokepoint for every read path — SSE reconnect/replay and the GET-history endpoints — so it covers them all. `created` has been `NOT NULL` since the first migration, so this retroactively corrects historical runs (no backfill needed).
3. **`views/agui.py`** — `drive_llm_stream` stamps `event.timestamp` from the server clock **before** the event is queued/persisted, so live clients get the server time too. Stamping is unconditional, matching the existing house pattern `run.finished = agui_util._timestamp()`.

### Design notes
- **Live == replay:** the live path persists the emit-time stamp into the event's `data`, so replay of new rows returns the exact value the live client saw (fill-only preserves it); legacy rows fall back to `created`.
- **No new column:** the event's own protocol `timestamp` field is the single home; a dedicated `occurred_at` column was a considered YAGNI cut (nothing queries events by time at the SQL level).
- **Asymmetry is intentional:** live overwrites (backend owns emit time), replay preserves (read-time backfill). They can't diverge because the only writer of `data["timestamp"]` is the live path.

## Testing
- TDD throughout. New tests for the ms conversion (incl. naive→UTC), replay fill-from-`created`, fill-only preserve, and live stamping with real-clock bounds. Existing round-trip tests updated to assert the timestamp is populated while comparing all other fields unchanged.
- `uv run pytest` → **3589 passed, 100.00% branch coverage**; `ruff check` / `ruff format --check` clean.
- Reviewed by code-quality, test-coverage, comment-accuracy, and silent-failure agents; findings addressed.

## Acceptance criteria
- [x] Replayed events carry a `timestamp` reflecting per-event creation time (incl. historical runs)
- [x] Live events carry a `timestamp` reflecting server emit time, present on the wire
- [x] No AG-UI protocol or generated-schema changes
- [x] 100% branch coverage; lint/format clean

### DIFF
--- a/tests/unit/agui/test_agui_persistence.py
+++ b/tests/unit/agui/test_agui_persistence.py
@@ -1107,7 +1107,8 @@ async def test_threadstorage_save_run_events(
         strict=True,
     ):
         assert found_event == exp_event
-        assert db_event == exp_event
+        assert isinstance(db_event.timestamp, int)
+        assert db_event.model_copy(update={"timestamp": None}) == exp_event
 
 
 @pytest.mark.asyncio
@@ -1162,9 +1163,11 @@ async def test_threadstorage_save_single_event(the_async_session):
 
     assert len(pairs) == 2
     assert pairs[0][0] == 0
-    assert pairs[0][1] == event_0
+    assert isinstance(pairs[0][1].timestamp, int)
+    assert pairs[0][1].model_copy(update={"timestamp": None}) == event_0
     assert pairs[1][0] == 1
-    assert pairs[1][1] == event_1
+    assert isinstance(pairs[1][1].timestamp, int)
+    assert pairs[1][1].model_copy(update={"timestamp": None}) == event_1
 
     # Query with after_index=0 should only return the second
     pairs_after = await ts.list_run_events_after(
@@ -1177,7 +1180,8 @@ async def test_threadstorage_save_single_event(the_async_session):
 
     assert len(pairs_after) == 1
     assert pairs_after[0][0] == 1
-    assert pairs_after[0][1] == event_1
+    assert isinstance(pairs_after[0][1].timestamp, int)
+    assert pairs_after[0][1].model_copy(update={"timestamp": None}) == event_1
 
 
 @pytest.mark.asyncio
--- a/tests/unit/agui/test_agui_schema.py
+++ b/tests/unit/agui/test_agui_schema.py
@@ -1,11 +1,13 @@
 from unittest import mock
 
 import pytest
+from ag_ui import core as agui_core
 from sqlalchemy import orm as sqla_orm
 from sqlalchemy.ext import asyncio as sqla_asyncio
 
 from soliplex import util
 from soliplex.agui import schema as agui_schema
+from soliplex.agui import util as agui_util
 from soliplex.config import installation as config_installation
 from tests.unit.agui import agui_constants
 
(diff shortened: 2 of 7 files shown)


tests/unit/agui/test_agui_util.py
```diff
@@ -22,3 +25,14 @@ def test__timestamp(dt, tz):
     assert found is dt.now.return_value
 
     dt.now.assert_called_once_with(tz.utc)
+
+
+@pytest.mark.parametrize(
+    "value, expected",
+    [
+        (datetime.datetime(1970, 1, 1, 0, 0, 1, tzinfo=datetime.UTC), 1000),
+        (datetime.datetime(1970, 1, 1, 0, 0, 1), 1000),
```
tseaver, line 34: Needs a test for a non-UTC timezone, too, as a guard against an implementation bug which blindly replaces the timezone.

tests/unit/agui/test_agui_schema.py
```diff
@@ -197,6 +199,51 @@ def test_runevent_from_agui_event(agui_event):
...
+        session.add(thread)
+        session.commit()
+
+        run = agui_schema.Run(thread=thread)
+        session.add(run)
+        session.commit()
+
+        run_event = agui_schema.RunEvent.from_agui_model(
+            run,
+            agui_core.events.RawEvent(event={"k": "v"}, timestamp=999),
```
tseaver, line 237: Please make the timestamp value a named, module-scope constant.

tests/unit/agui/test_agui_schema.py
```diff
@@ -335,7 +382,9 @@ async def test_run_events(the_session):
             await run.list_events(),
             strict=True,
         ):
-            assert db_event == agui_event
+            assert isinstance(db_event.timestamp, int)
```
tseaver, line 385: The tests for `to_agui_model` cover this part. Just comment on why the strip:
```suggestion
            # `db_event.timestamp` gets overwritten: strip it.
```

tests/unit/agui/test_agui_persistence.py
```diff
@@ -1107,7 +1107,8 @@ async def test_threadstorage_save_run_events(
         strict=True,
     ):
         assert found_event == exp_event
-        assert db_event == exp_event
```
tseaver, line 1110: Add a module-scope helper function for this repeated comparison, e.g.
```python
def _compare_events_without_timestamp(db_event, exp_event):
    assert db_event.model_copy(updaete={"timestamp": None}) == exp_event

tseaver (commented):
I'm not sure exactly what the purpose is here.  The `pydantic_ai.ui.agui._event_stream` has been setting timestamps on the AG-UI events since back in December.

91jaeminjo:
Closing this. After digging into the actual runtime, @tseaver was right: the AG-UI events emitted from Pydantic AI already carry a per-event `timestamp`, so this PR's premise — that per-event timestamps aren't on the wire for live or replay — doesn't hold.

**Backend (verified against `main` + the real persistence DB):**
- Pydantic AI's `AGUIEventStream.handle_event` fill-stamps `timestamp` (ms since epoch) on every event at generation time, and `_events_to_activity` stamps skill sub-agent events. `main` never maps `run.created` onto `event.timestamp`.
- In the actual thread DB, **460/461 persisted events already have a per-event `timestamp`**, and within a single run those values are genuinely per-event (e.g. one run has 11 distinct timestamps across 22 events, spanning run-start → run-finish). The only null is one `TOOL_CALL_RESULT`.

**Frontend (verified in `soliplex_frontend`):**
- The chat-timestamp feature resolves each message's time as `event.timestamp ?? runCreated` (`agui_event_processor.dart`), where `runCreated` = the run's `created` and is only a **fallback** (null while live). It already gets accurate per-event times today, with or without this PR.

**Why not merge anyway:**
- The **replay** change (fill from `created`) is a no-op for every event that already carries a timestamp.
- The **live** change is an *unconditional* overwrite that replaces Pydantic's generation-time stamp with a dequeue-time stamp; for compacted delta messages it shifts the time from message-start to message-end — a small regression. (The PR's own plan specified fill-only; the merged code diverged to unconditional overwrite.)

**The one real gap** that remains: events Pydantic synthesizes on its *exception-cleanup* path (interrupted-tool-call results) bypass the stamping override and go out unstamped. It's rare and only user-visible on the live tile (replay falls back to `run.created`). Tracking it separately on `fix/agui-error-path-timestamp` with a minimal fill-only stamp, plus an upstream report to pydantic-ai — not this broad change.

Thanks @tseaver for the push-back.

PR #1140, opened 2026-07-14
Add per-room use_subagents skill config

ggozad:
Adds a `use_subagents` flag to a room's `skills:` config (`RoomSkillsConfig`), threaded into `SoliplexSkillToolset`:

```yaml
skills:
  installation_skill_names: ["my-skill"]
  use_subagents: false   # default: true
```

Default is true, so existing rooms are unchanged. Setting false exposes the skill's tools directly to the room's agent instead of running the skill as an isolated sub-agent.
This is useful for skills that need multiple turns to complete (e.g. a form-filling skill) when using small models.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ dependencies = [
     "fakeredis != 2.35.0",  # brownbag
     "greenlet",
     "haiku.rag-slim >= 0.66, < 0.67",
-    "haiku-skills >= 0.18.0, < 0.19",
+    "haiku-skills >= 0.18.1, < 0.19",
     "idna >= 3.15",
     "itsdangerous",
     "joserfc >= 1.6.7, < 1.7",  # CVE-2026-48990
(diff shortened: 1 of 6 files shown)


tseaver (commented):
How does the room agent know about the `SKILL.md` or other references / resources associated with the skill?

ggozad:
> How does the room agent know about the `SKILL.md` or other references / resources associated with the skill?

You are right @tseaver I should also have included the prompt builder flag.
Added `use_subagents` flag to `build_system_prompt`. The main agent's prompt now has the appropriate tools and prompt to call the skill's tools, read resources etc.

At some stage, we should spend some time together and figure out whether we really need haiku.skills or how we can simplify it. There is some work in pydantic AI that overlaps with what we do, the remaining gap is subagents and AGUI integration.

tseaver (approved):

tseaver (approved):

PR #1245, opened 2026-08-12
Register haiku.rag's evidence capabilities in room configuration

ggozad:
Configures haiku.rag's two evidence capabilities as room skills and pins `haiku.rag-slim`
to `>= 0.74.0, < 0.75`.

## Evidence skills

Two skill kinds, neither taking any configuration:

- `haiku.rag.skills.evidence_compaction`: on each request, replaces earlier questions'
  evidence with a capsule of what was cited, re-attaching cited page images. Requests only;
  stored history is untouched.
- `haiku.rag.skills.citation_policy`: requires every answer to declare what grounds it, or
  that nothing does. Publishes a `citation_policy` state namespace and schema.

Compaction without the citation policy loses evidence, because the capsule is built only
from what was cited: an uncited question's evidence becomes a receipt with nothing in its
place. Configure both, or neither. `docs/config/rooms.md` covers it, and the example rooms
that retrieve configure both.

They are skills rather than `agent.capabilities` entries because `models.Skill` publishes a
skill's name, description, state namespace and state schema to clients, while nothing in
`agent.capabilities` reaches the API at all.

## Why 0.74.0

Earlier releases read an absent prompt as a resumption, which is how pydantic-ai's UI adapter
runs every turn, so every AG-UI run raised before its first token. 0.74.0 settles the
question from a flag on the evidence record instead.
`tests/functional/test_ui_host_contract.py` covers that shape, which the unit tests cannot
reach because they mock both the adapter and the agent.

## Also

- The `analysis` room configured `haiku.rag.skills.rag` and `haiku.rag.skills.analysis` over
  one corpus, which gives the agent two near-identical search tools, splits its citations
  across two records and gives one question two search budgets. Analysis alone now.
- `rag_lancedb_override_path` expands a leading `~`.
- A room's `haiku.rag.yaml` was documented as requiring the `haiku_chat` agent kind, which
  does not exist.

## Verification

`uv run pytest tests/unit`: 3911 passed, 7 xfailed, 100% coverage. Ruff, format and
pymarkdown clean, and no new `ty` diagnostics.

Driven live against the ORB multimodal corpus (1000 documents, 121k chunks) with
Muse-Glimmer-30B, seven turns in one thread: turns 1, 2, 3 and 5 retrieved and cited; turn 4
answered a visual back-reference with no search at all, from the page image retained in the
capsule, and re-cited it; turns 6 and 7 refused with empty declarations. The thread ended
with no policy violations.


docs/config/agents.md
```diff
@@ -115,6 +115,51 @@ agent:
...
+  Register the citation policy alongside compaction, or register neither.
+
+- The citation policy without compaction is safe:  citations are required
+  and recorded, and no history is rewritten.
+
+A room that registers neither resends every earlier turn's full search
+results on every request.
+
+  ```yaml
+  agent:
```
tseaver, line 156: This seems to be an odd place to hang this configuration:  why is this not a flag on the `haiku.rag.skills.rag` / `"haiku.rag.skills.analysis` skill configs instead?

ggozad: Both capabilities are one per agent singletons, not one per skill: each discovers and merges the records of all the room's evidence capabilities. A room can legitimately hold two evidence skills over different corpora (say rag on one db and analysis on another), so a flag on each skill config would be two switches for one capability, with no defined meaning when they disagree.

The skills block would avoid that too, but these are agent-level Pydantic AI capabilities. One could potentially choose a different compaction mechanism for example, when running with a better model or say reduced context.

ggozad: You are right about the analysis room: it had rag and analysis over the same `rag_lancedb_stem`, which I now discourage in the haiku.rag docs as it duplicates search tools, split citations and forces two budgets. Fixed to analysis alone, and `docs/config/rooms.md` no longer shows the pairing.

ggozad:
I tested with both TUI as well as the Flutter client on a few of multi-turn sessions, and both compaction and citability capabilities work as expected.

ggozad:
Fixed a minor bug:
`rag_lancedb_override_path` never expanded `~`, it was joined to the room directory as a literal. It's now expanded before the join: absolute paths are unaffected, relative ones still resolve against the room config's directory.

docs/config/agents.md
ggozad: I have thought a bit more on this and I think I side now with what you were thinking:
I see skills as capabilities with config that are visible to the clients.
`models.Skill.from_config` publishes each one's name, description, state namespace and state schema to clients, while anything named in `agent.capabilities` appears nowhere in that payload. 
The citation policy owns a `citation_policy` namespace and records violations into it, so registering it as a bare capability meant no client could discover either the guarantee or the state behind it. Originally I didn't think this would be necessary, but it might be in the future.
I will refactor to use skills.

tseaver:
@ggozad you wrote in the description above:
> A client reopening a thread rebuilds its history from stored runs but has no state to send,
so the run would compact against an empty record. ThreadStorage.get_latest_state answers
with the state a run ended with, its final STATE_SNAPSHOT, taken along the lineage of the
run being started; the POST handler fills it in when the client sent none. A client's own
state always wins. No schema change: RunAgentInput already stores the whole input as JSON.

I'm not sure I believe that theory of operation:

- Each of the "stored runs" carries its initial AG-UI state plus its own AG-UI event stream, from which deriving the terminal AG-UI state is straightforward, and something the frontend already supports for "live" threads.

- When resuming a thread without the state from an immediately previous run, the client is supposed to use those two elements from the "parent" run to synthesize the "final" AG-UI state for the parent run, and pass that as the initial AG-UI state for the "child" run, just as they would do if the thread were not interrupted.

- [This frontend issue](https://github.com/soliplex/frontend/issues/474) and related (e.g. https://github.com/soliplex/frontend/issues/56) seem to indicate that the frontend team knows they are supposed to handle this correctly.

I'd like to see some evidence `;)` that the backend should change here, rather than the frontend.

ggozad:
> @ggozad you wrote in the description above:
> 
> > A client reopening a thread rebuilds its history from stored runs but has no state to send,
> > so the run would compact against an empty record. ThreadStorage.get_latest_state answers
> > with the state a run ended with, its final STATE_SNAPSHOT, taken along the lineage of the
> > run being started; the POST handler fills it in when the client sent none. A client's own
> > state always wins. No schema change: RunAgentInput already stores the whole input as JSON.
> 
> I'm not sure I believe that theory of operation:
> 
> * Each of the "stored runs" carries its initial AG-UI state plus its own AG-UI event stream, from which deriving the terminal AG-UI state is straightforward, and something the frontend already supports for "live" threads.
> * When resuming a thread without the state from an immediately previous run, the client is supposed to use those two elements from the "parent" run to synthesize the "final" AG-UI state for the parent run, and pass that as the initial AG-UI state for the "child" run, just as they would do if the thread were not interrupted.
> * [This frontend issue](https://github.com/soliplex/frontend/issues/474) and related (e.g. [Defense-in-depth: harden JSON Patch and send AG-UI state on every run frontend#56](https://github.com/soliplex/frontend/issues/56)) seem to indicate that the frontend team knows they are supposed to handle this correctly.
> 
> I'd like to see some evidence `;)` that the backend should change here, rather than the frontend.

You are right, I got carried over from my own haiku.rag client in the example app where the state handling is largely the server's responsibility.

I will drop the backend state-restoration and lineage changes from this PR, while retaining the final `STATE_SNAPSHOT` emission so clients always have an authoritative terminal state to replay.

tseaver:
@ggozad PR #1264, which I merged on Friday, added the `soliplex.agui.with_final_state` wrapper, which  the final `STATE_SNAPSHOT` to runs with `ERROR` status (`soliplex.views.agui.init_agent_stream` previously did that for `FINISHED`).

ggozad:
> @ggozad PR #1264, which I merged on Friday, added the `soliplex.agui.with_final_state` wrapper, which the final `STATE_SNAPSHOT` to runs with `ERROR` status (`soliplex.views.agui.init_agent_stream` previously did that for `FINISHED`).

Thanks for pointing it out, I will rebase

src/soliplex/config/skills.py
```diff
@@ -22,6 +24,9 @@
 from . import exceptions as config_exc
 from . import rag as config_rag
 
+if typing.TYPE_CHECKING:  # avoid an import cycle at runtime
+    from .installation import InstallationConfig
```
tseaver:
```suggestion
if typing.TYPE_CHECKING:  # avoid an import cycle at runtime
    from . import installation as config_installation
```

src/soliplex/config/skills.py
```diff
@@ -305,6 +310,107 @@ class HR_Analysis_SkillConfig(_HaikuRAGCapabilityConfig):
...
+    kind: typing.ClassVar[str]
+    capability_factory: typing.ClassVar[typing.Callable]
+    capability_id: typing.ClassVar[str]
+    name: typing.ClassVar[str]
+    description: typing.ClassVar[str]
+    source: typing.ClassVar[SkillKind] = SkillKind.NATIVE
+    state_namespace: typing.ClassVar[str | None] = None
+    state_type: typing.ClassVar[type[pydantic.BaseModel] | None] = None
+
+    _installation_config: InstallationConfig = None  # noqa F821 cycles
```
tseaver:
```suggestion
    _installation_config: config_installation.InstallationConfig = None  # noqa F821 cycles
```

src/soliplex/config/skills.py
```diff
@@ -305,6 +310,107 @@ class HR_Analysis_SkillConfig(_HaikuRAGCapabilityConfig):
...
+    _config_path: pathlib.Path | None = None
+
+    @classmethod
+    def from_yaml(
+        cls,
+        installation_config: InstallationConfig,  # noqa F821 cycles
```
tseaver:
```suggestion
        installation_config: config_installadion.InstallationConfig,  # noqa F821 cycles
```

src/soliplex/agents.py
```diff
@@ -119,11 +121,19 @@ def get_default_agent_from_configs(
...
-        if not isinstance(capability, cap_rag_audit.RAGAccessAuditCapability)
+        if not isinstance(
+            capability,
+            (
+                cap_rag_audit.RAGAccessAuditCapability,
+                haiku_compaction.EvidenceCompactionCapability,
+                haiku_policy.CitationPolicyCapability,
+            ),
+        )
     ]
```
tseaver: We need a better way to detect "hook-only" vs. "needs deferral" than listing them explicitly here.  I opened #1270 to track that.

tests/unit/config/test_config_skills.py
```diff
@@ -661,3 +661,93 @@ def test_entrypoint_capability_config_unknown_entry_point(
...
+            {"kind": klass.kind, "bogus": "nonesuch"},
+        )
+
+
+def test_every_skill_kind_advertises_registered_features(installation_config):
+    """A room's features are looked up by name when a thread is created.
+
+    An advertised name which is not in the registry is a KeyError there,
+    so every kind's names must be registered as this module is imported.
+    """
```
tseaver: This test needs to be tightened:  any skill config must either a) have either a non-null `state_namespace` and have `agui_feature_names` be a one-tuple, `(<state_namespace>`,), or b) have `state_namespace` set to `None` and `agui_feature_names` set to  (returning) `()`.  In case a), the `state_namespace` value should also be registered in `AGUI_FEATURES_BY_NAME` with the `model_klass` of the registration value set to the skill config class.

tests/unit/test_agents.py
```diff
@@ -310,6 +311,92 @@ def test_get_default_agent_aligns_rag_capability_vision(
...
+        capabilities=[routed],
+        rag_db_paths={},
+    )
+
+    agents.get_default_agent_from_configs(
+        agent_config=agent_config,
+        tool_configs={},
+        mcp_client_toolset_configs={},
+        capability_config=capability_config,
+    )
```
tseaver:
```suggestion
    found = agents.get_default_agent_from_configs(
        agent_config=agent_config,
        tool_configs={},
        mcp_client_toolset_configs={},
        capability_config=capability_config,
    )
    
    assert found is agent_klass.return_value
```

tests/functional/test_ui_host_contract.py
```diff
@@ -0,0 +1,99 @@
...
+
+pydantic-ai's UI adapter passes the whole conversation as
+``message_history`` and never passes a prompt.  haiku.rag's evidence
+capabilities settle which question a run is answering from that shape, so
+a promptless run is exactly the shape the unit tests cannot reach:  they
+mock both the adapter and the agent.
+
+``FunctionModel`` means these need neither an LLM nor an embedder.  The
+model answers with text and never calls a tool, so the RAG database is
+never opened.
```
tseaver, line 11: I need a better handle on what we are trying to exercise here:

tests/unit/test_agents.py
```diff
@@ -310,6 +311,92 @@ def test_get_default_agent_aligns_rag_capability_vision(
...
+    assert built == [rag_capability]
+
+
+@mock.patch("soliplex.config.agents.get_model_from_config")
+@mock.patch("pydantic_ai.Agent")
+def test_get_default_agent_keeps_citation_policy_out_of_routing(
+    agent_klass,
+    gmfc,
+):
+    """The citation policy capability is hook-only.
```
tseaver: Both the testcase name and the docstring are misleading, as the test doesn't use the citation policy capability at all.

This test likely belongs with the work outlined in #1270, rather than here.

src/soliplex/config/skills.py
tseaver: 35deac6

src/soliplex/config/skills.py
tseaver: 35deac6

tests/unit/config/test_config_skills.py
tseaver: 0a943a2

tseaver (commented):
@ggozad I've resolved all my comments (and pushed changes), except for my question about the new functest.

tests/functional/test_ui_host_contract.py
tseaver: I opened #1271 to track the question.

PR #1247, opened 2026-08-12
fix windows coverage errors

bd-mkt:
# fix: make the test suites pass on Windows as well as Linux

## Summary

The test suites only ran on Linux. On Windows the unit suite reported
**4 failures and 57 errors**, and the functional suite a further 4
errors.

Every one of those was a portability bug in the tests themselves. No
production code path under test was at fault, and no source file is
modified here: the failures were assertions and fixtures that had
quietly baked in POSIX assumptions.

Adds a `windows-latest` CI job so this does not regress.

## Root causes

Four distinct classes, in descending order of blast radius.

### 1. Windows paths expanded as escapes by `re.sub` (57 errors)

`tests/unit/cli/conftest.py` rewrites a scratch copy of
`example/minimal.yaml` to repoint its authz database. The replacement
was passed to `subn()` as a *string*, and `re.sub` expands backslash
escapes in replacement strings -- so a `tmp_path` of `C:\Users\...`
raised `re.PatternError: bad escape \U`. Now passed as a callable,
which bypasses escape processing entirely.

### 2. Windows paths inside a double-quoted YAML scalar

The same class of bug one layer down, and the more dangerous of the
two. With the regex fixed, the injected
`sync: "sqlite:///C:\Users\..."` then met YAML's own escape rules,
where `\U` requires eight hex digits.

Worth flagging for reviewers: `\U` failed loudly only by luck. `\t`,
`\L`, and `\N` are all valid YAML escapes and would have **silently**
corrupted the path instead.

Fixed by spelling these paths POSIX-style through a shared
`_sqlite_dburi()` helper. SQLAlchemy's sqlite dialect accepts
`sqlite:///C:/...` on Windows -- verified before relying on it.

`tests/unit/config/test_config_rooms.py` had the same defect:
`pathlib.Path("/tmp/rag.lancedb")` stringifies to `\tmp\rag.lancedb` on
Windows, so `\t` became a literal tab inside the fixture YAML. It now
interpolates `.as_posix()`, which round-trips to the same
`pathlib.Path` on either platform.

### 3. CRLF in captured subprocess output

`tests/unit/capabilities/test_filesystem_cap.py` compared script stdout
against an exact string. A child Python newline-translates `print` to
`\r\n` on Windows. The assertion now normalizes line endings.

`run_script` returns the script's stdout verbatim, which is correct
behaviour, so the fix belongs in the test rather than the capability.

### 4. POSIX mode bits

`tests/unit/skills/test_bwrap_sandbox.py` asserted
`st_mode & 0o777 == 0o600`. Windows has no POSIX mode bits -- `chmod`
there only toggles the read-only flag -- so a writable file reports
`0o666` however it was chmod'ed. The expected value is now derived from
`os.name`.

### Functional suite

`tests/functional/test_cli_smoketest.py` duplicates the scratch-
installation fixture and carried an identical copy of causes 1 and 2 (4
errors). Fixed the same way. All 21 non-`needs_llm` functional tests
now pass on Windows.

## Why not `skipif`

`addopts` includes `--cov=tests/unit` alongside
`--cov-fail-under=100`, so a test skipped on one platform leaves its
own body uncovered and fails the gate. Every fix here is therefore
platform-agnostic -- the same code runs on both -- rather than a skip.

## Verification

Run on both platforms, `-n auto`:

| | Linux | Windows |
| ---------------------- | ----------- | ----------- |
| Unit suite             | 3729 passed | 3729 passed |
| Coverage gate          | 100%        | 100%        |
| Functional (non-LLM)   | 4/4 changed | 21 passed   |

Linux was verified in a `python:3.13` container. `ruff check`,
`ruff format --check`, `actionlint`, `pymarkdown`, and `check-toml` all
pass on the changed files.

### DIFF
--- a/tests/functional/test_cli_smoketest.py
+++ b/tests/functional/test_cli_smoketest.py
@@ -50,6 +50,18 @@ STALE_ACL_MARKER = "stale_filter_func"
 ROOM = "faux"
 
 
+def _sqlite_dburi(db_path: pathlib.Path, driver: str = "") -> str:
+    """Build an absolute 'sqlite:///<abs>' URI for 'db_path'.
+
+    The path is spelled POSIX-style because these URIs are written into a
+    double-quoted YAML scalar, where a Windows path's backslashes are
+    escape sequences: 'C:\\Users\\...' fails to parse ('\\U' wants eight
+    hex digits) and e.g. '\\t' would silently become a tab. SQLAlchemy's
+    sqlite dialect accepts 'sqlite:///C:/...' on Windows.
+    """
+    return f"sqlite{driver}:///{db_path.as_posix()}"
+
+
 @pytest.fixture
 def scratch_installation(tmp_path):
     """A copy of 'example/minimal.yaml' backed by a fresh, empty authz DB."""
@@ -59,10 +71,15 @@ def scratch_installation(tmp_path):
     db_path = tmp_path / "authz.sqlite"
 
     text = config_path.read_text()
-    text, n_db = _AUTHZ_DBURI_RE.subn(
+    # Passed as a callable because 're.sub' expands backslash escapes in a
+    # replacement *string*, which a Windows 'tmp_path' would trip over.
+    authz_replacement = (
         "authorization_dburi:\n"
-        f'  sync: "sqlite:///{db_path}"\n'
-        f'  async: "sqlite+aiosqlite:///{db_path}"',
+        f'  sync: "{_sqlite_dburi(db_path)}"\n'
+        f'  async: "{_sqlite_dburi(db_path, "+aiosqlite")}"'
+    )
+    text, n_db = _AUTHZ_DBURI_RE.subn(
+        lambda _: authz_replacement,
         text,
     )
     assert n_db == 1, f"expected one authz_dburi stanza, found {n_db}"
@@ -162,7 +179,7 @@ def _seed_stale_acl_entry(db_path, room_id):
     'list_room_policies' read.
     """
     session = authz_schema.get_session(
-        engine_url=f"sqlite:///{db_path}",
+        engine_url=_sqlite_dburi(db_path),
         init_schema=True,
     )
     with session:
--- a/tests/functional/test_cli_smoketest.py
+++ b/tests/functional/test_cli_smoketest.py
@@ -21,6 +21,7 @@ import yaml
 
 from soliplex import authz
 from soliplex.authz import schema as authz_schema
+from tests._dburi import sqlite_dburi
 
 # 'tests/functional/test_cli_smoketest.py' -> parents[2] is the repo root.
 _REPO_ROOT = pathlib.Path(__file__).resolve().parents[2]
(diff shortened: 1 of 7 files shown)


tests/unit/capabilities/test_filesystem_cap.py
```diff
@@ -234,7 +234,9 @@ async def test_run_script_returns_stdout(temp_dir):
         "scripts/echo.py", "stanza jabberwocky --stanza 3"
     )
 
-    assert output == "stanza jabberwocky --stanza 3\n"
+    # 'run_script' hands back the script's stdout verbatim, and a child
+    # Python on Windows newline-translates 'print' to '\r\n'.
+    assert output.replace("\r\n", "\n") == "stanza jabberwocky --stanza 3\n"
```
tseaver, line 242: Are you running Python 3.12 on the Windows machine?  AFACT, for Python >= 3.13, the default is already to read in "universal newlines" mode, which should normalize `"\r\n"` to `"\n"`.

https://docs.python.org/3/library/pathlib.html#reading-and-writing-files

tests/functional/test_cli_smoketest.py
```diff
@@ -50,6 +50,18 @@
...
+def _sqlite_dburi(db_path: pathlib.Path, driver: str = "") -> str:
+    """Build an absolute 'sqlite:///<abs>' URI for 'db_path'.
+
+    The path is spelled POSIX-style because these URIs are written into a
+    double-quoted YAML scalar, where a Windows path's backslashes are
+    escape sequences: 'C:\\Users\\...' fails to parse ('\\U' wants eight
+    hex digits) and e.g. '\\t' would silently become a tab. SQLAlchemy's
+    sqlite dialect accepts 'sqlite:///C:/...' on Windows.
+    """
+    return f"sqlite{driver}:///{db_path.as_posix()}"
```
tseaver: Could this bit be replaced with a pair of fixtures defined in `tests/unit/conftest.py`, `authz_dburi_sync` and `authz_dburi_async`?

tests/unit/skills/test_bwrap_sandbox.py
```diff
@@ -18,6 +19,11 @@
 RUN_ID = uuid.uuid4()
 USERNAME = "phreddy"
 
+# 'write_transcript' requests owner-only '0600'. Windows has no POSIX mode
+# bits -- 'os.chmod' there only toggles the read-only flag -- so a writable
+# file reports '0666' however it was chmod'ed.
+EXPECTED_TRANSCRIPT_MODE = 0o600 if os.name == "posix" else 0o666
```
tseaver: Might need tweakage for macos?  I'm pretty sure HFS supports the group / world settings.

tests/unit/config/test_config_rooms.py
```diff
@@ -87,7 +91,7 @@
 skills:
     skill_configs:
         - kind: "haiku.rag.skills.rag"
-          rag_lancedb_override_path: "{LANCE_DB_OVERRIDE_PATH}"
+          rag_lancedb_override_path: "{LANCE_DB_OVERRIDE_YAML_PATH}"
```
tseaver, line 94: Could that be a fixture as well in `tests/conftest.py`?

tseaver (changes requested):

tests/unit/capabilities/test_filesystem_cap.py
bd-mkt: No — 3.13.11 (Python 3.13.11 (main, Dec 9 2025) [MSC v.1944 64 bit (AMD64)]).

The universal-newlines default you're describing is correct, but it's a text-mode feature, and this output never passes through a text wrapper. run_script doesn't read a file — it decodes bytes captured from a subprocess pipe:


completed = await anyio.run_process([sys.executable, str(target), *argv], ...)
parts = [completed.stdout.decode("utf-8", errors="replace")]
anyio.run_process has no text / encoding / universal_newlines parameter — I checked the signature; it returns bytes only. And bytes.decode() is a pure codec call: it does no newline translation. The \r\n is already in the bytes, because the child Python's text-mode stdout put it there.

On 3.13.11, Windows:


1. bytes on the pipe      : b'hello\r\n'
2. .decode('utf-8')       : 'hello\r\n'   <-- what run_script does; CRLF survives
3. Path.read_text()       : 'hello\n'     <-- the pathlib path you linked; normalized
4. subprocess(text=True)  : 'hello\n'     <-- also normalized, but anyio has no text mode

Rows 3 and 4 are the universal-newlines machinery working exactly as you describe. Row 2 is the one in play.

tseaver: Sorry if misread that.

tseaver (approved):

tseaver (approved):

PR #1248, opened 2026-08-12
force UTF-8 not host locale

bd-mkt:
# fix: read and write text files as UTF-8, not the host locale encoding

## Scope

Stand-alone PR: `fix/cp1252_encoding` -> `main`, one commit
(`abc952c8`), 13 files. `main` is a direct ancestor, so it merges
without a prerequisite.

Five of those test files carry uncommitted follow-up edits (comments and
one assertion message only -- no logic change); amend them into
`abc952c8` or add a fixup before opening the PR.

Deliberately **not** included, to keep this reviewable on its own:

- The Windows test-portability fixes (a separate branch). Those repair
  tests that fail on Windows for unrelated reasons -- see "Verification"
  for why that matters when reading this PR's Windows results.
- `pyproject.toml` / `AGENTS.md` / the `windows-latest` CI job, which
  belong with that other change.

## The bug

`pathlib.read_text()`, `write_text()`, and `Path.open()` in text mode
fall back to `locale.getpreferredencoding(False)` when given no
`encoding=`. That is UTF-8 on Linux but `cp1252` on a typical Windows
host, so the same file decodes differently depending on where the code
runs.

YAML 1.2 and JSON (RFC 8259) both mandate UTF-8. The encoding of these
files is guaranteed by their format; the code was substituting the
encoding of whatever machine happened to run it.

The common cases fail **silently**. A UTF-8 file containing
`Ada Lovelace’s — terse.` reads back as `Ada Lovelaceâ€™s â€” terse.`
with no exception raised. Only bytes landing in one of cp1252's five
undefined slots (`0x81`, `0x8D`, `0x8F`, `0x90`, `0x9D`) raise at all.
On the write side the asymmetry is worse: Latin-1-range characters are
silently written as cp1252, while CJK or emoji raise
`UnicodeEncodeError`.

Practical consequences on a Windows host: corrupted system prompts fed
to the LLM, corrupted quiz question text, wrong values from `.env` and
from file-sourced secrets (a mojibaked secret authenticates against
nothing and gives no clue why), and `as-yaml` exports that are not
valid UTF-8 for the Linux server or for git.

## The fix

`encoding="utf-8"` at all 13 text-mode sites. No behaviour change on
Linux, where the fallback was already UTF-8.

| Area | Sites |
| ----------------------------------------- | ----- |
| `config/installation.py` (YAML, dotenv)   | 2 |
| `config/agents.py` (`prompt.txt`)         | 1 |
| `config/quizzes.py` (question JSON)       | 1 |
| `secrets.py` (file-sourced secret)        | 1 |
| `cli/admin_users.py`, `cli/room_authz.py` | 4 |
| `cli/audit.py` (logging config)           | 1 |
| `util.py` (git hash / branch / tag)       | 3 |

All YAML loading funnels through `_load_config_yaml`, so the single
`installation.py:107` change covers `installation.yaml`, every
`room_config.yaml`, completions, quizzes, and OIDC configs.

The two `tui/rest_api.py` sites open `"rb"` and were already immune.

### Behavioural (4)

Covering the paths that carry prose or credentials: installation YAML,
`prompt.txt`, quiz question JSON, and file-sourced secrets. Each writes
its fixture with `write_bytes(...)`, so the file on disk is UTF-8
regardless of the host default, and each fixture mixes both failure
modes -- `’` / `—` mojibake silently, while `Ł` (U+0141) raises.

All four were confirmed to **fail against the pre-fix source** and pass
after it, so they are not vacuous.

One subtlety: `json.dumps` escapes non-ASCII to `\uXXXX` by default, so
a Python-written question file is accidentally immune. The quiz test
writes `ensure_ascii=False` to exercise the decode path as a human- or
tool-authored file would.

### Guard against reintroduction (`tests/unit/test_encoding.py`, new)

The behavioural tests have a structural limit: on Linux the buggy code
is *correct*, so they can only fail where the locale encoding is not
UTF-8. On Linux CI they are vacuous.

So this adds an AST guard that walks every module under `src/soliplex/`
and fails on any text-mode call lacking `encoding=` -- rejecting the
pattern rather than waiting for a mojibaked byte to reach an assertion.
It reports exactly 13 violations against the pre-fix source and none
now. A second test feeds it a synthetic module containing every
offending form, so it cannot pass merely by finding nothing.

Two alternatives were evaluated and rejected as the primary mechanism:

- `ruff PLW1514` catches only 2 of the 13 -- it flags `Path.open()` but
  not `read_text()` / `write_text()`.
- `PYTHONWARNDEFAULTENCODING=1` (PEP 597) catches all four APIs and
  would be airtight against the 100% coverage gate, but
  `EncodingWarning` is attributed to the *caller*, and the test suite
  uses these APIs heavily -- roughly 100+ further sites under `tests/`.
  Left as an optional follow-up; the guard buys the same protection for
  `src/` without that churn.

## Verification

**Linux -- green.** 3735 passed, 100.00% coverage (23477 statements, 0
missed, 3424 branches, 0 partial), in a `python:3.13` container. This
is the platform CI runs on today, so it is the gating result. Measured
on a tree carrying this commit plus the separate portability fixes; a
branch-only re-run is in flight and those five extra files are
Linux-neutral.

**Windows -- the 6 new tests pass; the suite does not.** Running the
full unit suite on Windows from this branch gives:

| | Unit suite result |
| ------------------------- | ------------------------------------- |
| `main` (before this PR)   | 4 failed, 3668 passed, 57 errors |
| this branch               | 4 failed, 3674 passed, 57 errors |

The failure set is **identical** -- same 4 failures, same 57 errors --
with exactly the 6 new tests added to the passing count. Those 61
failures are pre-existing on `main` and are unrelated to encoding
(regex/YAML escaping of Windows paths, CRLF, POSIX mode bits); they are
fixed on the separate portability branch. This PR neither causes nor
resolves them.

Run in isolation, the 6 tests added here pass on Windows.

`ruff check` and `ruff format --check` pass on all changed files.


tests/unit/config/test_config__utils.py
```diff
@@ -58,6 +58,23 @@ def test__load_config_yaml_w_invalid(temp_dir, invalid):
...
+NON_ASCII_PROSE = "Ada Lovelace’s notes — Łódź"
+
+
+def test__load_config_yaml_w_non_ascii(temp_dir):
+    """A UTF-8 config decodes the same whatever the host locale is."""
+    config_path = temp_dir / "config.yaml"
+    yaml_text = f'id: "{NON_ASCII_PROSE}"\n'
+    config_path.write_bytes(yaml_text.encode("utf-8"))
+
+    found = config_installation._load_config_yaml(config_path)
```
tseaver: This testcase belongs in `tests/unit/config/test_config_installation.py`:  it is exercising a change in the correspondingly-named module (`src/soliplex/config/installation.py`).

src/soliplex/util.py
```diff
@@ -67,17 +67,17 @@ def __init__(self, file_path: str | pathlib.Path):
         hash_path = self.repo_dir / "git-hash.txt"
 
         if hash_path.is_file():
-            self._git_hash = hash_path.read_text().strip()
+            self._git_hash = hash_path.read_text(encoding="utf-8").strip()
```
tseaver, line 71: Changes in this module need a corresponding test or tests in `tests/unit/test_util.py`.

tests/unit/test_encoding.py
```diff
@@ -0,0 +1,130 @@
...
+in 'config/test_config__utils.py', secrets in 'test_secrets.py'), but
+those can only fail on a host whose locale encoding is not UTF-8. This
+guard is the part with teeth on every platform: it rejects the *pattern*
+rather than waiting for a mojibaked byte to reach an assertion.
+
+'ruff PLW1514' covers only the 'open()' form -- not 'read_text()' /
+'write_text()' -- which is why this is a test rather than a lint rule.
+It deliberately scans 'src/' only; the test suite's own ~100 call sites
+are a separate cleanup.
+"""
```
tseaver: This *has* to be a lint-time thing:  it is unacceptable for a check like this to fail due to adding another (even correct `path.read_text(encoding="utf-8")` call elsewhere in the codebase.  If wee need to write a script which runs at pre-commit / CI time to check for these cases, that script cannot embed the fragile counts below.

tseaver (changes requested):

tseaver (approved):

PR #1289, opened 2026-08-21
synchronize between skill.md and code

bd-mkt:
# fix: align `bubble-sandbox` tool docs with the actual tool signatures

## Summary

Every tool signature documented in the `bubble-sandbox` skill's `SKILL.md`
was wrong, and `list_volume_files` returned values that matched neither its
own docstring nor `SKILL.md`. Because `SKILL.md` is injected verbatim as the
agent's instructions (`_instructions()`), the agent was being told to call
these tools with parameter names that do not exist.

This PR corrects the documentation to match the code, fixes the one place
where the code was genuinely wrong, and removes two stale references left
behind by an earlier rename.

### `src/soliplex/skills/bwrap_sandbox/__init__.py`

- **`list_volume_files` now returns usable paths.** It previously returned
  `sub.absolute().name`, i.e. bare basenames, while both its own
  description and `SKILL.md` promised absolute paths. It now returns the
  path the sandbox sees — `/sandbox/volumes/<volume>/<name>` — via a new
  `SANDBOX_VOLUMES_PATH` constant, so results can be passed straight to
  `run` / `run_python`. Results are `sorted()` for deterministic output.
  Note this deliberately does *not* return `str(sub.absolute())`: the host
  path is meaningless inside the sandbox and should not be leaked into the
  prompt.
- **Dropped the `EnvironmentInfo = dict[str, str]` fossil** in favour of
  the real `list[bs_models.EnvironmentInfo]`. The upstream model already
  matches what `SKILL.md` documents, field for field, so there is nothing
  to translate; serialization to the agent is unchanged (tool returns go
  through a generic `TypeAdapter[Any]`), but the derived `return_schema` is
  now correct rather than claiming `dependencies` is a string.
- **`execute` → `run`** in `RUN_DESCRIPTION` and `RUN_PYTHON_DESCRIPTION`.
- **Rewrote `RUN_DESCRIPTION`**, which described a general-purpose dev
  shell rather than this sandbox — it advertised "tests, builds, git
  commands, package installs", suggested `cd /project && make test`, and
  warned about `rm -rf` / `drop table`. It now documents the string-vs-list
  `command` forms, `environment_name`, `timeout`, and the real path rules
  (read-only volumes, write only under `/sandbox/work/`), and points at
  `run_python` for anything beyond a quick file peek — which is what
  `SKILL.md` already said.
- **`RUN_PYTHON_DESCRIPTION`**: documented the previously unmentioned
  `environment_name` and `timeout` parameters and the read-only volumes.
- **Implicit-optional annotations** `environment_name: str = None` and
  `timeout: float = None` → `str | None` / `float | None`, so the generated
  JSON schema stops declaring a non-nullable type with a null default.
- Typos: `buuble_sandbox` → `bubble_sandbox`, `thses` → `these`.

### `src/soliplex/skills/bwrap_sandbox/SKILL.md`

- Corrected all four tool signatures, including the string-vs-list
  `command` form and the optional `environment_name` / `timeout`, neither
  of which was mentioned at all.
- Added an explicit instruction to always pass `environment_name` using a
  name returned by `list_environments`. When `allowed_environments` filters
  out the configured `default_environment`, an omitted `environment_name`
  silently runs in an environment the agent was never shown; passing it
  explicitly closes that gap.
- `list_volume_files` is now described as *returning* a list rather than
  *printing* lines, with the corrected path values.
- Frontmatter `name: bwrap-sandbox` → `bubble-sandbox`, matching
  `CAPABILITY_NAME` and the config `kind:`.
- Scoped the "persistent working directory" claim to the run: `get_workdir`
  keys the directory on `run_id`, so `/sandbox/work` does not survive
  across runs in a thread.
- Fixed the worked example: it used a non-existent environment name
  (`default` vs the configured `bare`), `script_text=`, and was missing a
  closing paren.
- Typos: "one or mor files", "sandbox/volumes/thread" (missing leading
  `/`), "is a already stated", "The task is value" → "vague", "absolue",
  "run a an", "inside a the script call", "e.g,".

### `tests/unit/skills/test_bwrap_sandbox.py`

- Updated `test_skill_list_volume_files` expectations to the sandbox paths.

## Behaviour change

One, and it is the point of the PR: `list_volume_files` returns
`/sandbox/volumes/thread/orders.csv` where it previously returned
`orders.csv`. Any prompt or caller that re-joined the basename onto a
volume root itself should be checked — within this repo nothing does; the
value flows only to the agent, which now receives a path it can actually
open.

The `EnvironmentInfo` annotation change is behaviour-neutral today.
`return_schema` is only sent to the model when `include_return_schema`
resolves true, which defaults to false unless the
`IncludeToolReturnSchemas` capability is in use — so this fixes a latent
bug rather than an active one.

### DIFF
--- a/src/soliplex/config/skills.py
+++ b/src/soliplex/config/skills.py
@@ -379,7 +379,7 @@ class BwrapSandboxSkillConfig:
     _config_path: pathlib.Path | None = None
 
     id: str | None = None
-    default_environment: str = "bare"
+    default_environment: str = "default"
     allowed_environments: bwrap_sandbox.AllowedEnvironments = None
     sandbox_config: bs_config.Config = None
     volumes: bs_models.VolumeMap = _default_dict_field()
--- a/src/soliplex/skills/bwrap_sandbox/SKILL.md
+++ b/src/soliplex/skills/bwrap_sandbox/SKILL.md
@@ -1,9 +1,10 @@
 ---
-name: bwrap-sandbox
+name: bubble-sandbox
 description: |
     This skill runs Python in a bubblewrap
-    sandbox. Each run has a persistent working directory and read-only
-    access to uploaded files mounted under '/sandbox/volumes'.
+    sandbox. Each run gets a scratch working directory that lasts for the
+    duration of that run, plus read-only access to uploaded files mounted
+    under '/sandbox/volumes'.
 ---
 
 # Sandbox
@@ -15,17 +16,17 @@ Use it to compute results from files.
 
 Use the sandbox if **any** of these is true:
 
-- The task references one or mor files uploaded
-  under `sandbox/volumes/thread` or `sandbox/volumes/room`.
+- The task references one or more files uploaded
+  under `/sandbox/volumes/thread` or `/sandbox/volumes/room`.
 - The task asks for a number, count, table, or other value derived from data.
 - You were about to state a computed result without actually computing it.
 
 Do NOT use the sandbox if **any** of these is true:
 
 - The question is about definitions, explanations, or concepts.
-- The answer is a already stated in the conversation.
+- The answer is already stated in the conversation.
 - The task is to write code for the user to run, not to execute code yourself.
-- The task is value ("analyze the files", "take a look")
+- The task is vague ("analyze the files", "take a look")
   with no concrete question. Ask the user what they want to know
   before running anything.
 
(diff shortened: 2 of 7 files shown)


src/soliplex/config/skills.py
```diff
@@ -379,7 +379,7 @@ class BwrapSandboxSkillConfig:
     _config_path: pathlib.Path | None = None
 
     id: str | None = None
-    default_environment: str = "bare"
+    default_environment: str = "default"
```
tseaver, line 382: This change makes no sense with the `soliplex` package:  it has no `default` environment.  Downstream that wants a different default *must* configure it explicitly.

tseaver (commented):
I've ported the rest of this PR's branch to a new one, with some further tweaks, in PR #1299.

tseaver:
Superseded by PR #1299.  Marking as "draft" / "do not merge" to preserve this branch until downstream quits using it.

PR #1299, opened 2026-08-24
fix: 'skills/bwrap_sandbox/SKILL.md' is source of truth for skill properties

### LINKED ISSUE #1298: `soliplex.skills.bwrap_sandbox`: drift between `SKILL.md` and `__init__.py`
PR #1289 (fe04d0e) tries to address this drift, but the remaining commits in the PR introduce a non-existing `default_environment` (replacing `bare` with `default`).

- [ ] The `SKILL.md` `name:` metadata does not match `skills.bwrap_sandbxo.CAPABILITY_NAME`
- [ ] The `SKILL.md` `description:` metadata does not match `skills.bwrap_sandbxo.CAPABILITY_DESCRIPTION`
- [ ] The `SKILL.md` instructions do not always prefix upload files with `/sandbox`
- [ ] Bunches of typos around argument names.

tseaver:
- Parse skill properties using 'skills_ref' into 'SKILL_PROPERTIES'

  - Actual 'kind' for 'config.skills.BWrapSandboxSkillConfig' now derived from 'SKILL_PROPERTIES.name';  The old kind, "bubble-sandbox", is preserved as a non-deprecated alias.

  - Replace module-scope 'CAPABILITY_NAME' with 'SKILL_PROPERTIES.name' (looked up from parsed 'SKILL.md').

  - Replace module-scope 'CAPABILITY_DESCRIPTION' with 'SKILL_PROPERTIES.description' (looked up from parsed 'SKILL.md').

  - Replace module-scope alias, 'EnvironmentInfo', with explicitly- imported 'bubble_sandbox.models.EnvironmentInfo'.

  - Use 'SANDBOX_VOLUMES_PATH' in Python for all mentions of '/sandbox/volumes' (looked up from parsed 'SKILL.md').

  - Use 'SANDBOX_WORKDIR_PATH' in Python for all mentions of '/sandbox/work' (looked up from parsed 'SKILL.md').

  - 'list_volume_files' skill tool now returns absolute paths relative to the sandbox root, and sorts tehm.

  - Tool arguments 'environment_name' / 'timeout' are now explicily nullable ('pydantic_ai' ignores the default 'None' when building the JSON schema for the tool signature).  Instructions for the tool specify passing

  - Tweak instructions to clarify calling conventions, return values, and other tool usage.

  - Various typo fixes.

deps: 'skills-ref'

  - Parse 'SKILL.md' files.

Closes #1298.

Supersedes PR #1289.


bd-mkt:
The reason for changing "bare" to "default" is that agents in some circumstances will pick up  the environment="bare" statement in SKILL.md and use that as an indication that's what's expected even if there is no bare environment listed. If the other changes fix that issue then it's not necessary

tseaver:
@bd-mkt 
> The reason for changing "bare" to "default" is that agents in some circumstances will pick up the environment="bare" statement in SKILL.md and use that as an indication that's what's expected even if there is no bare environment listed. If the other changes fix that issue then it's not necessary

The place to change the `default_environment` downstream is in the individual room configs:  changing the default-default here to an environment name that doesn't even exist in this repo breaks any deployment which doesn't also have a `default` environment.

tseaver:
Test failure is tracked in issue #1300, for which I have a fix pending.

PR #1305, opened 2026-08-25
Explicit defer loading

### LINKED ISSUE #1297: Explicit config for "routing" capability deferral
PRs #1161, #1245  and #1275 established distinction between "routing" capabilities (e.g. `haiku.rag.capabilities.RAGCapability`, `soliplex.skills.bwrap_sandbox.SandboxCapability`) and "hook" capabilities (e.g., `haiku.rag.capabilities.policy.CitationPolicyCapability`, `soliplex.capabilities.rag_audit.RAGAccessAuditCapability`), and added a policy that a room with more than one "routing" capability defers loading all "routing" capabilities.

The intent of the policy is to avoid crowding the room agent's context with every capability / skill's instructions, in a room with an arbitrary number of such capabilties / skills.  However, this policy leads to unexpected behavior in rooms which configure e.g. both `haiku.rag.capabilities.RAGCapability` and `soliplex.skills.bwrap_sandbox.SandboxCapability`, leading operatators to attempt workarounds like trying to force a capability to be loaded using the room prompt.

ISTM we should replace the policy with explicit configuration: each entry in a room configuration's `skills.skill_configs` list can set an explicit `defer_loading` flag, which defaults to `True`:  this value is just passed on to the underlying `pydantic_ai.capabilities.*` constructor.

ggozad:
Replaces the policy which derived capability deferral from how many
"routing" capabilities a room configured.

Deferral is now per-entry configuration. Every `skills.skill_configs`
entry takes `defer_loading`, passed straight to the capability
constructor, and `installation_skill_names` entries may be a mapping to
carry the same flag:

```yaml
skills:
  installation_skill_names:
    - "bare-bones"
    - name: "stanzas"
      defer_loading: false

  skill_configs:
      rag_lancedb_stem: "rag"
      defer_loading: true
```

## Defaults

| Kind | `defer_loading` |
| --- | --- |
| `haiku.rag.skills.rag`, `haiku.rag.skills.analysis` | `false` |
| `bwrap_sandbox` | `false` |
| `entrypoint` | `true` |
| filesystem skills, via `installation_skill_names` | `true` |

## Hook-only capabilities

`_is_routing_capability` is kept, no longer to choose a policy but to
enforce an invariant: a capability offering the model no tools or
instructions is never deferred. 

Closes #1297

### DIFF
--- a/docs/config/agents.md
+++ b/docs/config/agents.md
@@ -121,7 +121,13 @@ agent:
   any registered via the
   [`meta.agent_capability_types`](meta.md) stanza.  Capabilities which
   need configuration, or which a room should advertise, are configured as
-  [skills](rooms.md#skill-configuration) instead.
+  [skills](rooms.md#skill-configuration) instead.  `kwargs` reach the
+  capability's constructor, so a model-facing capability's
+  [deferred loading](rooms.md#deferred-loading) is set there as
+  `defer_loading`.  A capability offering the model no tools or
+  instructions is always loaded eagerly, whatever its `kwargs` ask for:
+  there would be nothing to load, and its hooks would not fire until it
+  was.
 
 ### Example Ollama Configuration
 
--- a/docs/config/rooms.md
+++ b/docs/config/rooms.md
@@ -161,8 +161,9 @@ for configuring an agent.
 
 ### Skill Configuration
 
-- `installation_skill_names` (a list of strings, default empty);  if set,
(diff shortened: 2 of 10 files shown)


tseaver (approved):
@ggozad I made `defer_loading` explicit in all the `as_yaml` properties, so that users dont need to know / care about the defaults once they are dumped.

Issue #1319, opened 2026-08-27
Agent guesses namespaced tool names (`bwrap-sandbox:run_python`) when the system prompt names the skill


tseaver:
Downstream report against the newly-released `soliplex` v0.78.

The `bwrap_sandbox` hardening from #1309 is working -- the agent does end up
calling `list_environments` and running successfully -- but the agent makes an
incorrect tool call *first*, and burns several model round-trips before it
recovers.

From a Logfire trace of the run (transcribed from a screenshot; the times are
offsets from the start of the span):

```text
0.3s   bwrap-sandbox:list_environments
0.6s   list_environments
5.1s   bwrap-sandbox:run_python
10.8s  run_python
16.2s  run_python
16.7s  load_capability
17.2s  load_capability
19.1s  run
```

Two things stand out:

1. The agent calls `bwrap-sandbox:list_environments` and
   `bwrap-sandbox:run_python` -- namespaced tool names, which the sandbox
   toolset does not register. Note also the hyphen, where the skill's own name
   is `bwrap_sandbox`.

2. `load_capability` is only reached at 16.7s, after five earlier tool calls,
   and is called twice. The first call that can actually have done any work is
   the `run` at 19.1s.

Every event is green in Logfire, so nothing here surfaced as an error.

tseaver:
## Analysis

### The `bwrap-` prefix is not generated anywhere in our stack

I checked every place that could join a namespace onto a tool name:

| Source | Separator / shape |
| --- | --- |
| `PrefixedToolset` (backs `PrefixTools`, `MCPToolset.prefixed()`) | `prefix_toolname` -- underscore (`pydantic_ai/toolsets/prefixed.py:32`) |
| Native MCP tool names | `':'.join([MCPServerTool.kind, server_label])` -> `mcp_server:<label>` (`pydantic_ai/models/anthropic.py:3230`, `pydantic_ai/models/openai.py:5212`) |
| Tool-execution span names | `running tool` (v2) / `execute_tool <name>` (`pydantic_ai/_instrumentation.py:594`) |
| Logfire MCP integration | `MCP request: tools/call <name>` |
| Soliplex | nothing -- no `tool_prefix`, no `PrefixTools`, no `_` -> `-` rewriting anywhere in `src/soliplex/` |

`bwrap-sandbox` is not even our spelling of the skill:
`bwrap_sandbox.SKILL_PROPERTIES.name` is taken verbatim from the `SKILL.md`
[...]

PR #1321, opened 2026-08-28
chore: strip down 'bwrap_sandbox' room prompt

tseaver:
... and related skill instructions

- Avoid naming the 'bwrap_sanbox' skill, which is eagerly loaded as of 'soliplex 0.78'. Instead, focus on the tools which the skill makes available to the room agent.

- Update the 'bwrap_sanbox' skill's own instructions to avoid naming the sandbox skill:  they now describe using the tools it provides, and explicitly prohibit any namespacing.

All this in service of removing a distracting, emphasized literal name ('bwrap-sandbox' / 'bubble-sanbdbox'), which can confuse the agent into trying to call the tools through a non-existent namespace (the behavior reported in https://github.com/soliplex/soliplex/issues/1319).

Toward #1319.

### DIFF
--- a/example/rooms/bwrap_sandbox/prompt.txt
+++ b/example/rooms/bwrap_sandbox/prompt.txt
@@ -1,32 +1,51 @@
-You are a data analysis assistant. You use a bubblewrap sandbox to
-read files, run Python code, and answer questions about the user's
-data. You also have access to a document knowledge base for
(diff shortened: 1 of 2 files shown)


tseaver:
Unable to validate the hypothesis on which this change depends (see #1319, and also https://github.com/soliplex/lab_bench/issues/4).

PR #1322, opened 2026-08-31
feat: rooms can name several LanceDB databases

ggozad:
A room's RAG skill or tool can name several LanceDB databases and search
them as one, or name none and read the databases its `haiku.rag`
configuration places. Results, documents and audit records report which
database they came from.

Requires `haiku.rag-slim >= 0.82.0, < 0.83` (bumped in the first commit).

## Configuration

`rag_databases` replaces the single `rag_lancedb_stem` /
`rag_lancedb_override_path` pair; the two forms are mutually exclusive.
A configuration may also use neither, and read what its `haiku.rag`
configuration places.

```yaml
skills:
  skill_configs:
      rag_databases:
        - name: "papers"
          rag_lancedb_stem: "papers"
        - name: "wiki"
          rag_lancedb_override_path: "../wiki.lancedb"
```

Entries resolve their paths as the single-database fields do. Names must
be unique.

## Placement

A configuration naming no database of its own reads the ones its
`haiku.rag.yaml` places in `lancedb.databases`:

```yaml
lancedb:
  databases:
    archive: "s3://bucket/lancedb/archive.lancedb"
```

This is the only way for a room to read a database that is not a local
directory. Previously such a room had to name a local database it never
used, as a placeholder, and set `lancedb.uri` beside it.

**Breaking.** That placeholder arrangement no longer works, and now says
so. `haiku.rag` 0.82 places every database in `lancedb.databases`, and a
configuration carrying `lancedb.uri` fails to load with the replacement
spelled out and its own location echoed back:

```
lancedb
  Value error, lancedb.uri was removed; write lancedb.databases:
  {NAME: 's3://bucket/lancedb/example.lancedb'} instead
```

A room reaches this whether the `uri` is its own or the installation's,
since it reads the two merged, so an installation-wide `lancedb.uri`
fails every room. `soliplex-cli audit` reports it before a room is used,
in place of the unresolved path it would otherwise blame.

Naming a database in soliplex while that configuration already places
one in `lancedb.databases` is two placements for one room, and raises
`RagDbTwoPlacements` when the room is used. soliplex writes the
databases a room names into the client configuration, so without this it
would silently replace what the room configured and open the
placeholder.

This replaces silent breakage rather than introducing it: from
`haiku.rag` 0.79 a location soliplex passes wins over a configured
`lancedb.uri`, so these rooms already read the placeholder database and
returned nothing.

## Behaviour

Every database a config names is resolved into
`config.lancedb.databases`, including the one a stem or override path
names, which takes the file's stem for its name. A config naming none
leaves that configuration untouched, and reads the names out of it
through `DatabaseScope.resolve`. The client, the
capability and the audit CLI read the location from the config rather
than a `db_path` argument. A config naming one database is read as one
database, not as a set of one, so instructions and result formatting are
unchanged for existing rooms.

The name reaches the agent as the collection a result came from, the room
endpoints as `database`, and audit records as `name=location` pairs.
Single-database rooms gain the `name=` prefix in their audit records.
Locations stay in the configuration. Following #1324,
`extra_parameters` reports `database_names`, now for several databases as
well as one.

Candidates are combined by the configured reranker. Without one they are
ordered by cosine similarity to the query, which compares across
databases because a set shares an embedding model. Full-text searches
order by retrieval score.

The chunk endpoint asks each covered database in turn, since chunk IDs
repeat between copies of a database.

`MultipleRAGDatabasesProtocol` and `_MultiRAGDatabasesBase` are removed:
no implementer, and as structural protocols the `isinstance` dispatch
could never reach them. That dispatch also held an unreachable
`NameError`.

## Unrelated fix (severable)

`GET /rooms/{id}/search` returned 500 for any document without a title,
because `SearchHit` required strings where `haiku.rag` returns
`str | None`. Any corpus ingested without `auto_title` hits it.
`document_title` and `document_uri` are now nullable in the
`SearchResults` schema. Not breaking, since those responses were 500s,
but the Flutter repo should regenerate if it generates from OpenAPI.

## Testing

Unit tests throughout, 100% coverage maintained. Verified end to end
against two real databases: `soliplex-cli audit` reports the federated
document count, `/documents` and `/search` attribute each result,
`/chunk/{id}` renders page images for a chunk held only by the second
database, and the agent answered questions from each corpus in turn. A
single-database room written in the old form was checked to keep its
document count; its audit records gain the `name=` prefix.

Placement is covered by unit tests rather than end to end: a config
naming nothing reports the names, audit string and `database_names` its
`haiku.rag` configuration implies, including the default database; and
the conflict raises for both the stem and the `rag_databases` forms.

## Not included

- Selecting among a room's databases at query time. A room searches
  every database its configuration places. `haiku.rag` narrows a client
  with `sources=[...]` and raises for a name outside the room's set, so
  the follow-up is carrying a selection from the client through to the
  query. The room model reports the names today
  (`extra_parameters.database_names`), which is what a picker needs to
  offer.
- No example room: one corpus under two names returns every chunk twice
  and raises `AmbiguousCitationError`, and a real example needs two
  corpora the repo does not ship. The shape is documented instead.
- Duplicate skill names remain silently accepted
  (`extract_skill_configs` keys on name, so two differently configured
  sandboxes collapse to one). Pre-existing, orthogonal.
- `docs/config/rooms.md` still lists `ask` / `research` /
  `list_documents` / `get_document` as RAG skill tools; the capability
  exposes `rag_search` and `rag_cite`. Pre-existing drift.

### DIFF
--- a/src/soliplex/config/rag.py
+++ b/src/soliplex/config/rag.py
@@ -168,23 +168,22 @@ class _RAGConfigBase:
     ) -> hr_config.AppConfig:
         """Name this config's databases in 'lancedb.databases'
 
-        'rag_databases' comes from '_RAGDatabaseBase', which DB-bearing
-        configs mix in alongside this class.  haiku.rag treats 'lancedb.uri'
-        and 'lancedb.databases' as mutually exclusive, so naming databases
-        clears the URI.  Copies rather than mutates: the installation's own
-        config object is shared with every other room.
+        'rag_lancedb_databases' comes from '_RAGDatabaseBase', which
+        DB-bearing configs mix in alongside this class.  haiku.rag treats
+        'lancedb.uri' and 'lancedb.databases' as mutually exclusive, so
+        naming databases clears the URI.  Copies rather than mutates: the
+        installation's own config object is shared with every other room.
         """
-        rag_databases = getattr(self, "rag_databases", None)
+        databases = getattr(self, "rag_lancedb_databases", None)
 
-        if not rag_databases:
+        if not databases:
             return room_hr_config
 
         lancedb = room_hr_config.lancedb.model_copy(
             update={
                 "uri": "",
                 "databases": {
-                    entry.name: str(entry.rag_lancedb_path)
-                    for entry in rag_databases
+                    name: str(path) for name, path in databases.items()
                 },
             },
         )
@@ -270,6 +269,36 @@ class _RAGDatabaseBase:
             return rspdb
 
     @property
+    def database_name(self) -> str:
+        """The name this database answers to, marked when it is missing.
+
+        A config written with a stem or an override path names no database,
+        so the file's stem is its name.
+        """
+        try:
+            return self.rag_lancedb_path.stem
+        except RagDbFileNotFound as exc:
+            return f"MISSING: {exc.rag_db_filename.stem}"
+
+    @property
+    def rag_lancedb_databases(self) -> dict[str, pathlib.Path]:
+        """Every database this config names, keyed by name"""
+        if self.rag_databases:
+            return {
+                entry.name: entry.rag_lancedb_path
+                for entry in self.rag_databases
+            }
+
+        return {self.database_name: self.rag_lancedb_path}
+
+    @property
+    def rag_database_names(self) -> list[str]:
+        if self.rag_databases:
+            return [entry.database_name for entry in self.rag_databases]
+
+        return [self.database_name]
+
+    @property
     def rag_db_audit_path(self) -> str:
         """Identify the database(s) this config reads, for audit records"""
         if self.rag_databases:
@@ -281,27 +310,7 @@ class _RAGDatabaseBase:
         return str(self.rag_lancedb_path)
 
     def get_extra_parameters(self) -> dict:
-        if self.rag_databases:
-            return {
-                "rag_lancedb_paths": {
-                    entry.name: entry.get_extra_parameters()[
-                        "rag_lancedb_path"
-                    ]
-                    for entry in self.rag_databases
-                },
-            }
-
-        try:
-            rag_lancedb_path = self.rag_lancedb_path
-        except RagDbFileNotFound as exc:
-            missing = exc.rag_db_filename
-            db_name = f"MISSING: {missing.stem}"
-        else:
-            db_name = rag_lancedb_path.stem
-
-        return {
-            "database_names": [db_name],
-        }
+        return {"database_names": self.rag_database_names}
 
 
 def adjust_yaml_rag_databases(
@@ -338,6 +347,15 @@ class RAGDatabaseEntry(_RAGDatabaseBase):
     """
 
     name: str = None
+
+    @property
+    def database_name(self) -> str:
+        try:
+            self.rag_lancedb_path  # noqa B018
+        except RagDbFileNotFound:
+            return f"MISSING: {self.name}"
+
+        return self.name
 
     def __post_init__(self):
         if not self.name or not self.name.strip():
--- a/src/soliplex/config/rag.py
+++ b/src/soliplex/config/rag.py
@@ -44,6 +44,56 @@ class RagDbFileNotFound(ValueError):
         )
 
 
+class RagDbSingleAndMultiConflict(TypeError):
+    def __init__(self, _config_path):
+        self._config_path = _config_path
+        super().__init__(
+            f"Configure either 'rag_databases' or a single database via "
+            f"'rag_lancedb_stem' / 'rag_lancedb_override_path', not both "
+            f"(configured in {_config_path})"
+        )
+
+
+class RagDbDuplicateDatabaseName(TypeError):
+    def __init__(self, name, _config_path):
+        self.name = name
+        self._config_path = _config_path
+        super().__init__(
+            f"Duplicate name in 'rag_databases': {name!r} "
+            f"(configured in {_config_path})"
+        )
+
+
+class RagDbEntryRequiresName(TypeError):
+    def __init__(self, _config_path):
+        self._config_path = _config_path
+        super().__init__(
+            f"Each entry in 'rag_databases' needs a 'name' "
+            f"(configured in {_config_path})"
+        )
+
+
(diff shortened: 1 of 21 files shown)


src/soliplex/config/rag.py
```diff
@@ -110,30 +149,81 @@ def haiku_rag_config(self) -> hr_config.AppConfig:
                     room_config_yaml,
                 )
 
-                self._haiku_rag_config = hr_config.AppConfig.model_validate(
+                room_hr_config = hr_config.AppConfig.model_validate(
```
tseaver, line 199: Seems like we could just inject the single DB into HR config's `databases` as a one-entry list, and drop passing the explicit `database` argument to the client.

ggozad: Makes sense for `rag_lancedb_stem`, where the stem = name.

What would you use for `rag_lancedb_override_path: "../wiki.lancedb"`, which has no name in the config? Basename, or a fixed placeholder?

Asking because the key becomes `SearchHit.database` and the audit prefix, so a basename would make renaming the directory change the database's identity in the API and logs.

Otherwise agreed.

tseaver: basename seems right to me.  How does the RAGAudit stuff on `main` record it?

/me looks:

It uses the `self.db_paths` entry, which the agent sets from `capability_config.db_paths`.

tseaver:
/cc @91jaeminjo, @WilliamKarolDiCioccio 

The front-end is going to want two things from soliplex once this lands:
- The room model needs to contain (directly or indirectly) a list of the database names that are present by default in the room's rag sources:  these should be the logical names from the multi-db map.  For a non-multidb room, the field should be present, with a list containing the single logical name @ggozad and I were discussing above (the stem / basename).  If we simplify as discussed above, the value will always be just the keys of the haiku-rag config's `databases` field.
- For a room with multiple databases, the client will have a key available in the AG-UI state in which to pass a list of user-selected database names:  if that key is present, it should be used to narrow the databases in the HR config used to do the RAG query.  Passing a list which is not a subset of the room's list will raise an error.

ggozad:
> /cc @91jaeminjo, @WilliamKarolDiCioccio
> 
> The front-end is going to want two things from soliplex once this lands:
> 
> * For a room with multiple databases, the client will have a key available in the AG-UI state in which to pass a list of user-selected database names:  if that key is present, it should be used to narrow the databases in the HR config used to do the RAG query.  Passing a list which is not a subset of the room's list will raise an error.

Note that that key cannot change in the course of the conversation in the thread. Once set in the beginning it has to stay the same. i.e. you CANNOT filter the dbs after you start.

src/soliplex/config/rag.py
ggozad: Done, and rebased onto main so it picks up #1324.

Basename works out as `rag_lancedb_path.stem`, which is what #1324 already does for the single case, so both forms now name databases the same way and `get_extra_parameters` returns database_names for several exactly as it does for one. 

`db_path=` is gone from all four call sites (rooms dispatch, skill capability, search_documents, audit CLI)

On the audit question: the db_paths value is opaque, so I kept the bare path when the config uses rag_lancedb_stem / rag_lancedb_override_path, and name=path pairs only when it names a set. Existing rooms log exactly what they logged before. Happy to make it uniform if you would rather.

tseaver: I'd say let's treat the single-db forms as legacy; they are purely config sugar.  Under the covers, we turn them all into into a list of one item, including the way we query haiku-rag, how we show the DBs to the external models, and how we record them for audit.

ggozad: Every configuration now resolves into `lancedb.databases`, so the value is always that field's keys:

```
    room.skills["rag"].extra_parameters.database_names  ->  ["papers", "wiki"]
```

Same shape for the `search_documents` tool under `room.tools[...]`. The field is present for a single-database room too, reporting `["papers"]` from the file's stem, and a room that places its databases in its own
`haiku.rag.yaml` reports the keys it wrote there. Locations never appear, only names.

ggozad:
> * For a room with multiple databases, the client will have a key available in the AG-UI state in which to pass a list of user-selected database names:  if that key is present, it should be used to narrow the databases in the HR config used to do the RAG query.  Passing a list which is not a subset of the room's list will raise an error.

It is not (yet) possible to select among a room's dbs. They are always
what the room's configuration places. A follow up PR will have to do this.

src/soliplex/config/rag.py
```diff
@@ -110,42 +204,101 @@ def haiku_rag_config(self) -> hr_config.AppConfig:
...
+                },
+            },
+        )
+
+        return room_hr_config.model_copy(update={"lancedb": lancedb})
+
 
 @dataclasses.dataclass(kw_only=True)
 class _RAGDatabaseBase:
     """Base class for configs which expose a 'rag_lancedb_path' property"""
```
tseaver: Should likely be called `_RAGDatbasesBase`, and the docstring should referenc the `rag_databases` property.  I expected `rag_lancedb_path` to disappear entirely.

src/soliplex/config/rag.py
```diff
@@ -110,42 +204,101 @@ def haiku_rag_config(self) -> hr_config.AppConfig:
...
 
-    # One of these two options must be specified
+    # One of these two options must be specified, unless 'rag_databases'
+    # names several databases instead.
     rag_lancedb_stem: str = None
     rag_lancedb_override_path: str = None
```
tseaver: I hadn't expected that we would preserve these values as actual fields on the object:  I thought they would be declared as `dataclasses.InitVar[str]` (that was what I meant by "sugar only") and then merged into `self.rag_databases` in `__post_init__`.

tseaver: [bae8b02](https://github.com/soliplex/soliplex/pull/1322/commits/bae8b0280b44ceae2ef74fb37ad267638cd811b5)

src/soliplex/config/rag.py
tseaver: [bae8b02](https://github.com/soliplex/soliplex/pull/1322/commits/bae8b0280b44ceae2ef74fb37ad267638cd811b5)

PR #1347, opened 2026-09-08
docs: lead the RAG ingestion setup with a docling container

WilliamKarolDiCioccio:
Ingestion needs docling, which the server's `haiku.rag-slim` dependency
deliberately leaves out. Installing the full `haiku.rag` to get it costs
more than it looks: roughly 88 additional packages including `torch`,
`transformers`, `opencv` and the full NVIDIA CUDA stack, an upgrade to
`click` out from under the CLI, and two distributions left providing the
same import package.

Running docling as a container avoids all of that, and the compose
service listens on the port `example/haiku.rag.yaml` already points at,
so it needs no configuration change. This makes it the leading option
and documents what the pip route actually costs.

The CPU image is deliberate: the example corpus is Markdown, so there is
no OCR to accelerate and it avoids requiring the NVIDIA container
toolkit.

Split out of the context-metering work so it can land on its own.

https://claude.ai/code/session_01HGgKMJXcvXn53C8zjKxzfR

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -63,3 +63,33 @@ services:
       - ./db:/app/db
       - ./sandbox/workdirs:/app/sandbox/workdirs
       - ./uploads:/app/uploads
+
+  # Document conversion for RAG ingestion.
+  #
+  #   docker compose up -d docling_serve
+  #
+  # Only needed to BUILD a RAG database, never to serve one. The server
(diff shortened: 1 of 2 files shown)


tseaver:
Superseded by `soliplex-template v0.16`, which allows generating a full-service `compose` stack with `soliplex` cloned and running in dev mode.  See: https://soliplex.github.io/soliplex-template/getting-started/tutorials/backend-dev-mode/

PR #1348, opened 2026-09-08
feat(agui): report how full a thread's context window is

WilliamKarolDiCioccio:
Reports how full a thread's context window is, so a client can draw an
indicator without guessing.

**Depends on #1351** (`chore/haiku-rag-0.84`) for pydantic-ai 2.40+,
and is rebased onto it. GitHub will not let the base leave the #1347
stack via the API, so the diff shows #1351's two commits until #1351
merges; the branch itself no longer carries #1347.

## What is measured

`run_usage.final_input_tokens` records the provider's `prompt_tokens`
for the **last** model request of a run. That is the request as sent:
instructions, capability instructions, tool and MCP schemas, the chat
template, reasoning and tool calls the run carried, and RAG evidence as
compacted on the way out. `resolved_model_name` records which model
served it.

The existing `input_tokens` is cumulative across every request in a
run, so a tool loop reports several times the context actually sent.
The two answer different questions — what the run cost, and how full
the window was.

## Where the window comes from

`Model.context_window`, from the model's resolved profile. pydantic-ai
fills it for hosted models it knows (`gpt-4o` → 128000,
`gemini-2.5-flash` → 1048576) from `genai-prices`, with no I/O.

Ollama and any OpenAI-compatible provider behind a `base_url` resolve
to **no window**: pydantic-ai has nothing to look them up by. An agent
therefore gains one key:

```yaml
context_window: 32768
```

It is merged into the model's profile and overrides whatever
pydantic-ai resolved. Set it to what the runner is actually configured
to serve, not the nominal model size — Ollama defaults every model to a
far smaller window than the model advertises. Unset, a local model has
no window and the client shows no percentage.

`get_model_from_config()` no longer writes a dummy key into the
provider keywords it was handed; the Ollama provider is given the key
directly. The test that covered that branch only passed because it
asserted against the dictionary the code had just mutated.

## API

```text
GET /v1/rooms/{room_id}/agui/{thread_id}/context
```

```json
{
  "max_model_len": 32768,
  "model_name": "gpt-oss:latest",
  "measured_tokens": 244,
  "measured_at_run_id": "7b404155-…"
}
```

`max_model_len` is null when neither pydantic-ai nor the configuration
gives a window. `measured_tokens` is null until a run in the thread has
reached the model. A thread that does not exist or belongs to another
user is a 404, as its siblings are; `get_latest_measured_context()`
validates through `_find_user_thread()` first, so that is distinct from
a thread with nothing measured yet.

`AGUI_RunUsage` gains `final_input_tokens` and `resolved_model_name`.

## Withdrawn since the first revision

**The per-source breakdown (`?detail=true`)**, on review. It itemised
the transcript stored *before* run N while attributing run N's final
request, so everything the run itself added surfaced as overhead; and
its parts could sum past the total when compaction had shrunk the
request. Doing it correctly needs the final request and a tokenizer
to count it, which only vLLM offers.

**Provider probing and the vLLM compose service.** Superseded by the
profile. The direct `httpx` dependency goes with them.

**A factory agent** declares no model, so it reports no window. The
measurement still stands.

## Upgrading

Migration `a1c7d3e90b42` adds two nullable columns to `run_usage`.
**Required**: usage recording writes to them, so every run fails to
record on an un-migrated database.

```bash
uv run alembic -x soliplex.installation_path=<dir> upgrade head
```

No backfill, safe on a live database, `downgrade` drops both. Nothing
else changes meaning. Backend and frontend ship independently: a new
frontend against an old backend gets a 404 and shows no indicator.

## Testing

4221 tests, 100% branch coverage. The endpoint's tests build real
`AgentConfig` objects rather than mocks, so the window in each case is
what pydantic-ai actually resolves: a declared one, a hosted one it
knows, a hosted one overridden, and local and OpenAI-compatible ones it
does not. Live-tested against Ollama with the override set.

https://claude.ai/code/session_01Ccp5JZC2S5iDJybewL9ANL


WilliamKarolDiCioccio:
Hi @ggozad. I'm requesting your review first because you're the qualified one to determine if something was left out in the token usage computation. Only then I'm gonna ping Tres. Thx.

ggozad (commented):
Hey this is a great direction, some remarks which would simplify it quite a bit and generalize better:

#### Blocking

1.  Use pydantic-ai 2.42's model profile

We are moving to pydantic-ai 2.42 regardless of this PR. It already exposes the resolved context window:

```python
@property
def context_window(self) -> int | None:
    """The resolved profile's context_window."""
    return self.profile.get("context_window")
```

For custom and local models, `ModelProfile` explicitly supports:

```python
profile={"context_window": 128_000}
```

We already pass a partial `profile=` dictionary to `OpenAIChatModel` for Ollama and base-URL OpenAI-compatible providers through `_OPENAI_COMPAT_PROFILE` in `config/agents.py`. Add the configured override to that profile. When no override is supplied, pydantic-ai still fills known models from `genai-prices`.

This replaces `context_window.py` once the breakdown described below is removed. It also removes the new direct `httpx` dependency.

Do not add a vLLM development service to this repository. Ollama is our local runner specifically because it is multiplatform; introducing an NVIDIA-specific service to support this endpoint is the wrong direction.

2. Report a window for every supported provider

The endpoint currently reports a window only for vLLM. Soliplex supports Ollama and hosted providers as well, and Ollama is the default local runner.

Build the configured model with `get_model_from_config()` and read
`model.context_window`. Hosted and known models then receive profile data without provider-specific endpoint probing. Local Ollama models receive the explicit profile override when their served window differs from the nominal model value.

The existing helper is effectively no-I/O for this purpose, but it is not literally pure: the Ollama branch currently inserts a dummy API key into `agent_config.llm_provider_kw`. Avoid describing it as pure, and consider removing that mutation while touching this path.

3. Drop `?detail=true`

It itemizes the wrong messages. `add_run_input()` writes the client's POST body once, before the run streams. `run_agent_input.messages` is therefore the transcript before run N, while `measured_tokens` is run N's final request.

That final request also contains run N's thinking blocks, tool-call arguments, and tool results. They are included in the numerator but absent from the itemization, so they are mislabeled as `overhead`. On a RAG turn that can be the largest real slice.

`totals_by_kind()` can also sum beyond its own total:

```python
>>> cs.totals_by_kind(
...     [("userText", "a"), ("toolResult", "b")],
...     [500, 9000],
...     200,
... )
{"userText": 500, "toolResult": 9000}
```

Dropping a negative residual does not prevent an over-count. Compaction can shrink the measured request while the stored transcript still carries the full evidence text, leaving a stacked bar larger than `measured_tokens`. The docstring and PR description claim the opposite.

The implementation also makes one sequential HTTP request per stored piece, with a five-second timeout for each operation and no total deadline. Endpoint latency is therefore unbounded in the size of the transcript.

4. The unknown-thread test does not exercise real persistence

`test_get_room_agui_thread_id_context_unknown_thread` configures the mocked `get_latest_measured_context()` to raise `UnknownThread`, but the real method cannot do that. It runs a joined query and returns `None` when there is no matching row. Consequently, an unknown thread or a thread belonging to another user returns a 200 response with empty measurements rather than the
tested 404.

Either validate ownership through `_find_user_thread()` before querying, or change the endpoint contract and remove the impossible exception path and test. Given the surrounding AG-UI endpoints, a 404 is imo correct.

5. Test `collect_texts` against real AG-UI models

`tests/unit/test_context_segments.py` constructs only `Mock(spec_set=...)` objects, and the view tests mock both `count_tokens()` and `_get_run_input()`. The module whose job is interpreting `ag_ui.core.types` is therefore never tested against those types.

Real models expose another unsupported case: `UserMessage.content` is `str | list[InputContent]`, not always a string. `collect_texts()` passes a multimodal list through as tokenizer text despite claiming to return `list[tuple[str, str]]`. This would produce an invalid tokenizer request or a misleading attribution.

This test gap did not cause the temporal mismatch in blocker 3; that requires a test covering what is stored before a run versus what the final model request contains. Use real message models for the type contract and a run-level test for the timing contract.

### Non-blocking


- `context_window.py` is 212 lines, much of it rationale for rejected
  alternatives. Keep that history in the commit or PR description rather
  than in implementation comments.
- Tighten the PR description. Phrases such as “complete by construction,”
  “The win,” and “Bonus” add advocacy without clarifying the contract.
- Remove the claim that the breakdown always sums exactly to the measured
  total. The implementation has a tested branch in which it does not.

### TLDR

* Rebase the work onto the pydantic-ai 2.42 upgrade.
* Keep the two nullable usage fields, migration, and `_final_request_usage()`. Obtain the denominator from `Model.context_window`, with a configured profile override for local Ollama models. 
* Delete `context_window.py`, `context_segments.py`, the vLLM Compose service, and `?detail=true`. 
* Fix the unknown-thread contract. 
* Keep `resolved_model_name` only if the endpoint uses it to keep measurements tied to the model that produced them.

tseaver:
@WilliamKarolDiCioccio 
Please rebase this PR without being stacked on #1347, which was superseded by `soliplex-template v0.16`, which allows generating a full-service `compose` stack with `soliplex` cloned and running in dev mode.  See: https://soliplex.github.io/soliplex-template/getting-started/tutorials/backend-dev-mode/

PR #1350, opened 2026-09-08
portability fixes

bd-mkt:
# fix: portability of `_check_is_subdir`, and run the unit suite on Windows

## Summary

Twelve unit tests failed on Windows. Sorting them by *why* gives three
groups, and only the third is a genuine platform limit:

1. **A portability bug in production code** (1 test). `_check_is_subdir`
   rejected a path with an embedded NUL by relying on
   `PosixPath.resolve` raising `ValueError`. `WindowsPath.resolve` does
   not raise — it returns a name nothing can open, whose `.parent` *is*
   the expected parent, so the check passed the name straight through.
   **Fixed in the source**, not skipped.
2. **Assertions with a hard-coded separator** (3 tests). The code under
   test emits `str(pathlib.Path(...))`; the tests compared against a
   POSIX literal. Fixed by building the expectation through `pathlib`.
3. **Genuinely POSIX-only backends** (8 tests). An `O_NOFOLLOW` open, a
   FIFO, a symlink. These now probe the host, warn, and skip.

`uv run pytest` on Windows 11 goes from **12 failed / 4172 passed** to
**4176 passed, 8 skipped, exit 0**. Linux behaviour is unchanged: every
probe answers `True` there, so nothing skips and nothing warns, and the
one source change is verified equivalent (below).

## 1 — the portability fix (`skills/bwrap_sandbox/`)

`_check_is_subdir` guards the sandbox workdir tree: it is what stops a
room, thread, or run id from escaping its parent. It was written as

```python
try:
    resolved = path.resolve()
except ValueError as exc:            # only ever the embedded-NUL case
    raise InvalidSubdir(path, ep_resolved) from exc
else:
    if resolved.parent == ep_resolved:
        return

raise InvalidSubdir(path, ep_resolved)
```

The `except ValueError` is load-bearing — it is the *only* thing
rejecting a NUL — but it fires only because `PosixPath.resolve` happens
to raise. That is an implementation detail of one `pathlib` flavour, and
the check silently loses a rejection wherever it does not hold:

```text
Path("/tmp/x/\x00").resolve()
  PosixPath   -> ValueError: embedded null byte        -> rejected
  WindowsPath -> WindowsPath('C:/tmp/x/\x00')          -> ACCEPTED
                 .parent == the expected parent
```

Now the NUL is rejected on its own terms, and nothing depends on which
`pathlib` is loaded:

```python
if "\x00" not in str(path) and path.resolve().parent == ep_resolved:
    return

raise InvalidSubdir(path, ep_resolved)
```

Note this **replaces** the `try`/`except` rather than sitting in front of
it. Leaving both would make the `except` unreachable — dead code on every
platform, and a coverage-gate failure on Linux.

The bwrap sandbox only ever runs on Linux, so this is not a live
vulnerability. It is a check that was one `pathlib` implementation detail
away from being wrong, in code whose whole job is to reject a crafted
path; making it say what it means is worth the three lines regardless of
which platform runs it.

`src/soliplex/skills/bwrap_sandbox/__init__.py` goes from 99% to **100%**
on Windows, and the test runs everywhere instead of skipping.

### Equivalence check

Old and new implementations were run side by side under Linux (WSL) over
all 11 parametrized cases plus 11 further probes — a NUL in leading,
middle, and trailing position, a real symlink, a space, a tilde, a
glob character, a backslash, and a 200-character name:

```text
divergences: 0
```

On POSIX, `resolve()` raises `ValueError` only for the embedded NUL —
every other failure surfaces as `OSError`, which `resolve(strict=False)`
swallows — so the replaced `except` had exactly one trigger, and the new
condition covers it.

## 2 — path separators (3 tests, fixed, not skipped)

`RAGDatabaseEntry.as_yaml` and `rag_db_audit_path` render a parsed
`pathlib.Path` back out, so what they emit carries the host's separator:

```text
expected: papers=/data/papers.lancedb
found:    papers=\data\papers.lancedb
```

| Test | Fix |
| --- | --- |
| `test__rcb_deferred_databases_several` | build the audit line from `pathlib.Path("/data/papers.lancedb")` |
| `test__rde_yaml_roundtrip[None-../papers.lancedb]` | compare against `str(pathlib.Path(override))` |
| `test_sdtc_as_yaml[w_kw1]` | same, in the expected `rag_databases` entry |

These are not weakened assertions: on Linux each expression evaluates to
the exact literal it replaced (verified). The separator is the only thing
that moves.

## 3 — POSIX-only backends (8 tests, warn and skip)

New `tests/_platform.py` holds the probes and three helpers. A test that
needs one calls it; the helper emits an `UnsupportedPlatformWarning`
(pytest lists it in the warnings summary, naming the skipped test) and
skips:

| Test(s) | Needs | Failure on Windows |
| --- | --- | --- |
| `test__open_no_symlinks` (7 params) | `requires_o_nofollow()` | `AttributeError: module 'os' has no attribute 'O_NOFOLLOW'` |
| `test_get_workdirs_room_thread_run_only[…symlink]` | `requires_symlinks()` | `OSError: [WinError 1314] A required privilege is not held` |

All seven parameters of `test__open_no_symlinks` reach
`_open_no_symlinks`, whose whole purpose is the `O_NOFOLLOW` open, so the
guard sits at the top of the test rather than per parameter. The symlink
and FIFO guards sit at their creation sites, so only the parameter that
actually needs the capability skips.

Probes ask the host, not `os.name`, wherever they can. `HAVE_SYMLINKS`
creates one in a temporary directory and reports whether it worked, so a
Windows box with Developer Mode turned on runs the symlink tests rather
than skipping them.

`tests/_platform.py` deliberately lives in `tests/`, not `tests/unit/`.
The coverage targets include `tests/unit`, and a probe whose branches are
by definition unreachable on whichever platform is running could never
reach the 100% gate. `tests/_dburi.py` is the existing precedent.

## The coverage gate 

The remaining skips leave the code they cover unmeasured, so
`--cov-fail-under=100` fails on Windows however the tests themselves
fared. After the portability fix the entire residue is one function and
its tests:

```text
src\soliplex\views\sandbox_workdirs.py     75  14  14  0   80%   112-135
tests\unit\views\test_sandbox_workdirs.py 153  26  36  1   78%   113, 174, 234-263
TOTAL                                   26174  40 3760  1   99.80%
```

`src/soliplex/skills/bwrap_sandbox/__init__.py` no longer appears — the
portability fix closed it.

A green suite that still exits non-zero is not really "runs on Windows",
so `tests/conftest.py` gained a `pytest_configure` that drops the
threshold to 0 **only when `os.name != "posix"`**, and warns that it has:

> coverage is not gated on a non-POSIX host: the tests covering the
> POSIX-only sandbox paths skip here, so `--cov-fail-under` has been
> relaxed to 0 for this run. The 100% gate is enforced on Linux (and in
> CI); re-check coverage there before relying on it.

Linux and CI are untouched — the hook returns on its first line there. It
also returns under `--no-cov`, where the threshold is set but unenforced
and the warning would be pure noise. The trade is that a Windows
developer will not see a genuine coverage regression locally, which is
why the warning says so in as many words. **Drop this hook if you would
rather Windows users pass `--no-cov` by hand**; the rest of the PR stands
without it.

One implementation note worth a reviewer's eye: pytest-cov reads its
threshold from the namespace it captured in
`pytest_load_initial_conftests` (`early_config.known_args_namespace`),
*not* from `config.option`. Mutating `config.option.cov_fail_under` looks
right and does nothing.

## Provenance

None of this is drift from a later change — each test has failed on
Windows since it landed:

| Test | Landed in | Date |
| --- | --- | --- |
| `test__open_no_symlinks`, workdir symlink listing | `5dc95ea3` (#1332) | 2026-09-02 |
| `test__check_is_subdir[\x00]` | `df7372df` (#1328) | 2026-09-02 |
| the three path-separator assertions | `f1717b93` (#1322) | 2026-09-04 |

`df7372df` is the one where the failing test was pointing at a real
defect rather than at a platform limit.

## Changes

| File | Change |
| --- | --- |
| `src/soliplex/skills/bwrap_sandbox/__init__.py` | **portability fix** — reject an embedded NUL explicitly |
| `tests/_platform.py` | **new** — probes and three `requires_*` helpers |
| `tests/conftest.py` | `pytest_configure` relaxing the coverage gate off POSIX |
| `tests/unit/views/test_sandbox_workdirs.py` | 3 guard calls |
| `tests/unit/config/test_config_rag.py` | 2 assertions built through `pathlib` |
| `tests/unit/config/test_config_tools.py` | 1 assertion built through `pathlib` |
| `AGENTS.md` / `DEVELOPMENT.md` | document the convention, and when *not* to reach for a probe |

211 insertions, 10 deletions. The only change under `src/` is the
seven-line `_check_is_subdir` body.

## Verification

On Windows 11 (Python 3.13.11), where all twelve previously failed:

- `uv run pytest -n 8` — **4176 passed, 8 skipped, exit 0**; each skip
  prints its reason in the warnings summary
- `src/soliplex/skills/bwrap_sandbox/__init__.py` — **100%** (was 99%,
  missing 447-448)
- `uv run ruff check` / `ruff format --check` — clean, 183 files
- `uv run pre-commit run pymarkdown` on both edited docs — passed

On Linux, checked under WSL rather than assumed:

- old vs new `_check_is_subdir` over 22 inputs — **0 divergences**
- `HAVE_SYMLINKS`, `HAVE_MKFIFO`, `HAVE_O_NOFOLLOW` all report `True`
- every `requires_*` helper is a no-op under
  `warnings.simplefilter("error")` — nothing skips, nothing warns
- `str(pathlib.Path(s)) == s` holds for all three rewritten expectations

The full suite has not been re-run on Linux from this branch; the changes
above are what would differ, and each is verified to be a no-op there.

Also swept for sibling cases: `grep` across `tests/` for `symlink`,
`mkfifo`, `O_NOFOLLOW`, `geteuid`, `pwd`, `grp`, and `resource` turned up
no other POSIX-only usage beyond the files touched here. The one prior
accommodation (`EXPECTED_TRANSCRIPT_MODE`, keyed on `os.name == "nt"` in
`test_bwrap_sandbox.py`) already handled its own case and is left alone.

### DIFF
--- a/tests/_platform.py
+++ b/tests/_platform.py
@@ -1,95 +1,53 @@
-"""Platform capability probes for tests needing a POSIX backend.
+"""Platform gating for the tests of the POSIX-only sandbox.
 
-Soliplex sandboxes run on Linux, so a handful of unit tests exercise
-behaviour the standard library offers only on POSIX hosts: 'O_NOFOLLOW'
-opens, FIFOs, and symlinks. Nothing here changes what those tests
-assert -- on Linux they run exactly as they did. On a host which cannot
-express the precondition (Windows, most obviously) the test warns and
-skips, rather than failing on an 'AttributeError' or a 'WinError 1314'.
+Soliplex sandboxes a room's tool calls with bubblewrap, which exists
+only on Linux: 'src/soliplex/skills/bwrap_sandbox/' drives 'bwrap'
+itself, and 'src/soliplex/views/sandbox_workdirs.py' serves the workdir
+tree it leaves behind, guarding the walk with an 'O_NOFOLLOW' open.
 
-Each 'requires_*' helper is called from inside the test that needs it,
-so the warning names the skipped test instead of firing once at import
-time. 'HAVE_*' is the same answer as a plain boolean, for a test which
-must decide per parameter rather than skip outright.
+Their tests build the same preconditions the machinery does -- a
+symlink, a FIFO, a name with an embedded NUL that 'PosixPath.resolve'
+rejects -- so on a host offering none of them there is nothing left to
+assert. Rather than skip test by test, each module skips whole: mark it
+with 'requires_posix_sandbox' and name it (with the module it covers)
+in 'POSIX_ONLY_COVERAGE_OMIT', which 'tests/conftest.py' drops from the
+coverage report so the 100% gate keeps its meaning off POSIX.
 
 This module lives in 'tests/', not 'tests/unit/', on purpose: the
-coverage targets in 'pyproject.toml' include 'tests/unit', and a probe
-whose branches are by definition unreachable on whichever platform is
-running could never satisfy the 100% gate.
+coverage targets in 'pyproject.toml' include 'tests/unit', and a
+platform gate whose other branch is by definition unreachable on
+whichever host is running could never satisfy the 100% gate.
+'tests/_dburi.py' is the existing precedent.
 """
 
 import os
-import pathlib
-import tempfile
-import warnings
 
 import pytest
 
 
 class UnsupportedPlatformWarning(UserWarning):
-    """A test was skipped: the host cannot express its precondition."""
+    """Something was skipped: the host cannot express its precondition."""
 
 
-def _probe_symlinks() -> bool:
-    """Report whether this host lets the test user create a symlink
+HAVE_POSIX_SANDBOX = os.name == "posix"
 
-    Windows does, but only for an administrator or with Developer Mode
-    turned on, so 'os.name' alone does not answer it -- make one and
-    see.
-    """
-    with tempfile.TemporaryDirectory() as td:
-        temp_path = pathlib.Path(td)
-        target = temp_path / "probe-target"
-        target.write_text("probe", encoding="utf-8")
+POSIX_ONLY_REASON = (
+    "needs a POSIX host: the bubblewrap sandbox, and the workdir tree "
+    "it leaves behind, are Linux-only, and so are the symlinks, FIFOs "
+    "and NUL-rejecting path resolution these tests set up"
+)
 
-        try:
-            (temp_path / "probe-link").symlink_to(target)
-        except (OSError, NotImplementedError):
-            return False
-        else:
-            return True
+requires_posix_sandbox = pytest.mark.skipif(
+    not HAVE_POSIX_SANDBOX,
+    reason=POSIX_ONLY_REASON,
+)
 
-
-HAVE_SYMLINKS = _probe_symlinks()
-HAVE_MKFIFO = hasattr(os, "mkfifo")
-HAVE_O_NOFOLLOW = hasattr(os, "O_NOFOLLOW")
-
-
-def _warn_and_skip(reason: str) -> None:
-    warnings.warn(reason, UnsupportedPlatformWarning, stacklevel=3)
-    pytest.skip(reason)
-
-
-def requires_symlinks() -> None:
-    """Skip unless this host can create a symlink"""
-    if not HAVE_SYMLINKS:
-        _warn_and_skip(
-            "needs symlinks: this host refuses to create one (on Windows, "
-            "run as an administrator or enable Developer Mode)"
-        )
-
-
-def requires_mkfifo() -> None:
-    """Skip unless this host has 'os.mkfifo'"""
-    if not HAVE_MKFIFO:
-        _warn_and_skip("needs FIFOs: 'os.mkfifo' is POSIX-only")
-
-
-def requires_o_nofollow() -> None:
-    """Skip unless 'os.open' accepts the POSIX-only 'O_NOFOLLOW'"""
-    if not HAVE_O_NOFOLLOW:
-        _warn_and_skip("needs a no-follow open: 'os.O_NOFOLLOW' is POSIX-only")
-
-
-def requires_posix_paths() -> None:
-    """Skip unless 'pathlib' resolves paths the POSIX way
-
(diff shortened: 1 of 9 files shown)


tests/_platform.py
```diff
@@ -0,0 +1,95 @@
...
+            return True
+
+
+HAVE_SYMLINKS = _probe_symlinks()
+HAVE_MKFIFO = hasattr(os, "mkfifo")
+HAVE_O_NOFOLLOW = hasattr(os, "O_NOFOLLOW")
+
+
+def _warn_and_skip(reason: str) -> None:
+    warnings.warn(reason, UnsupportedPlatformWarning, stacklevel=3)
```
tseaver: Why emit a warning during testing?  The "skip" notification should be sufficient.

tests/unit/skills/test_bwrap_sandbox.py
```diff
@@ -564,6 +565,13 @@ async def test_skill_run_python_w_extra_args(
     ],
 )
 def test__check_is_subdir(temp_dir, subpath, expectation):
+    if "\x00" in subpath:
```
tseaver: Couldn't we just skip this whole testcase module for Windows?  It cannot actually run the sandbox, anyway.

tseaver (approved):
@bd-mkt 
I've pushed 7d8801f, which skips the two sandbox-related testcase modules entirely, versus piecemeal, including removing them and their modules-under-test from coverage.

I also pushed 47cfed8, which gets the unittests running on Windows (Python 3.13 only), so that you are no longer the "canary in the mine" for future Windows regressions.

PR #1351, opened 2026-09-10
Update to haiku.rag 0.84

ggozad:
## haiku.rag 0.84.0

Bumps `haiku.rag-slim` 0.82 → `>= 0.84.0, < 0.85` and the dependencies 0.84 requires: `fastmcp` 3.3 → 4.0.2 (MCP SDK 2), `pydantic-ai-slim`>= 2.40. Dev: `pytest-asyncio` 1.4, `ruff` 0.16.6 and the matching pre-commit rev. `mcp_auth`, `mcp_client` and `mcp_server` needed no
changes for FastMCP 4.

ruff 0.16 formats Python inside Markdown fences, which reflows `bwrap_sandbox/SKILL.md`; `*.md` is excluded with `force-exclude`.

### Visual grounding for multi-dbs

0.84 puts `source` on `HaikuRAG.visualize_chunk`, so the chunk visualization endpoint drops its `reader_for` hop.

`GET /v1/rooms/{room_id}/chunk/{chunk_id}` also takes an optional `database`. Chunk IDs can potentially repeat between copies of a database, so the scan it falls back to without one picks arbitrarily between copies and can render pages from the wrong database; a citation already carries the name.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -110,11 +110,9 @@ show_missing = true
 [tool.ruff]
 line-length = 79
 exclude = [
-    # ruff 0.16 formats Python in Markdown fences.  Skill instructions and
-    # docs are prose whose wording is measured, so ruff stays off them.
-    "*.md",
+    "*.md",  # ruff 0.16 formats Python in Markdown fences
 ]
-force-exclude = true  # hold the exclusion for explicitly named files too
+force-exclude = true
 target-version = "py313"
 
 [tool.ruff.lint]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,9 +60,9 @@ dev = [
     "coverage >= 7.13.5, < 7.14",
     "pytest >= 9.0.2, < 9.1",
     "pytest-cov >= 7.0.0, < 7.1",
-    "pytest-asyncio >= 1.4.0, < 1.5",
+    "pytest-asyncio >= 1.3.0, < 1.4",
     "pytest-xdist >= 3.8.0, < 3.9",
-    "ruff >= 0.16.6, < 0.17",
+    "ruff >= 0.15.6, < 0.16",
     "soliplex-skills >= 0.6",
 ]
 docs = [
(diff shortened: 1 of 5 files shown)


src/soliplex/views/rooms.py
```diff
@@ -223,22 +223,31 @@ async def get_room_documents(
...
         return await rag.get_chunk_by_id(chunk_id), rag.source
 
-    for database in rag.source_names:
-        chunk = await rag.get_chunk_by_id(chunk_id, source=database)
+    for covered in rag.source_names:
+        chunk = await rag.get_chunk_by_id(chunk_id, source=covered)
 
         if chunk:
-            return chunk, database
+            return chunk, covered
```
tseaver: This whole section seems a bit convoluted:  would it work just to loop over `rag.source_names` unconditionally, e.g.:
```suggestion
    for covered in rag.source_names:
        if database is None or covered == database:
            chunk = await rag.get_chunk_by_id(chunk_id, source=covered)

        if chunk:
            return chunk, covered

    return None, None
```

pyproject.toml
```diff
@@ -60,9 +60,9 @@ dev = [
     "coverage >= 7.13.5, < 7.14",
     "pytest >= 9.0.2, < 9.1",
     "pytest-cov >= 7.0.0, < 7.1",
-    "pytest-asyncio >= 1.3.0, < 1.4",
+    "pytest-asyncio >= 1.4.0, < 1.5",
     "pytest-xdist >= 3.8.0, < 3.9",
-    "ruff >= 0.15.6, < 0.16",
+    "ruff >= 0.16.6, < 0.17",
```
tseaver: Changing `ruff` versions seems unrelated to this PR.

I'm not opposed, but would like to have an issue spelling out the rationale and expected rule changes, and have that work done in its own PR:  we would likely want to do this across all repos in the organization at about the same time, so that developers reset their "must do it this way to keep ruff happy" mental models once.

src/soliplex/views/rooms.py
ggozad: Simplified, thanks for catching.

pyproject.toml
ggozad: Reverted back together with `pytest-asyncio`

tseaver: I opened #1359 to track this issue.

tseaver (approved):

PR #1355, opened 2026-09-11
feat(agui): report how full a thread's context window is

WilliamKarolDiCioccio:
Reports how full a thread's context window is, so a client can draw an
indicator without guessing.

Rebased on `main` (#1351 is in). Addresses @ggozad's review on #1348
and @tseaver's on this PR.

## What is measured

`run_usage.final_input_tokens` records the provider's `prompt_tokens`
for the **last** model request of a run. That is the request as sent:
instructions, capability instructions, tool and MCP schemas, the chat
template, reasoning and tool calls the run carried, and RAG evidence as
compacted on the way out. `resolved_model_name` records which model
served it.

The existing `input_tokens` is cumulative across every request in a
run, so a tool loop reports several times the context actually sent.
The two answer different questions — what the run cost, and how full
the window was.

## Where a client reads it

Per @tseaver, there is no dedicated endpoint. The two numbers live
where the client already looks:

**The window is on the room.** `DefaultAgent.context_window`, returned
by `GET /v1/rooms/{room_id}`. It is fixed for the life of the process,
so it is read once. Null when nothing knows it.

**The measurement is on each run.** `GET /v1/rooms/{room_id}/agui/{thread_id}`
now carries each run's `usage`, so a client opening a thread has every
run's `final_input_tokens` and shows the latest — falling back across a
run that recorded none (an error). For the run it has just driven,
`GET /v1/rooms/{room_id}/agui/{thread_id}/{run_id}/usage` returns the
`AGUI_RunUsage` alone, without the events; null when the run never
reached the model.

## Where the window comes from

`Model.context_window` on the model's resolved profile. pydantic-ai
fills it for hosted models it knows (`gpt-4o` → 128000,
`gemini-2.5-flash` → 1048576) from `genai-prices`, with no I/O.

Ollama and any OpenAI-compatible provider behind a `base_url` resolve
to **no window** — pydantic-ai has nothing to look them up by. An agent
therefore gains one key:

```yaml
context_window: 32768
```

It is merged into the model's profile and overrides whatever
pydantic-ai resolved. Set it to what the runner is actually configured
to serve, not the nominal model size — Ollama defaults every model to a
far smaller window than the model advertises. Unset, a local model has
no window and a client shows no percentage.

`get_context_window_from_config()` answers null rather than raising for
a configuration the model cannot be built from — an agent template with
no model name, or a provider whose key is not set. Both fail loudly when
a run starts; listing the room is not that moment.

`get_model_from_config()` no longer writes a dummy key into the
provider keywords it was handed; the Ollama provider is given the key
directly. The test that covered that branch only passed because it
asserted against the dictionary the code had just mutated.

## Withdrawn along the way

- **`GET .../{thread_id}/context`** and `AGUI_ThreadContext` — redundant
  with the above (@tseaver).
- **The per-source breakdown (`?detail=true`)** — it itemised the
  transcript stored *before* run N while attributing run N's final
  request, and could sum past its own total (@ggozad).
- **Provider probing, the vLLM compose service, the direct `httpx`
  dependency** — superseded by the profile (@ggozad).

## Upgrading

Migration `a1c7d3e90b42` adds two nullable columns to `run_usage`.
**Required**: usage recording writes to them, so every run fails to
record on an un-migrated database.

```bash
uv run alembic -x soliplex.installation_path=<dir> upgrade head
```

No backfill, safe on a live database, `downgrade` drops both. Nothing
else changes meaning: `DefaultAgent` and `AGUI_RunUsage` gain optional
fields, the thread listing's runs gain `usage`, one endpoint is added.

## Testing

4442 tests, 100% branch coverage. `get_context_window_from_config` is
tested against real `AgentConfig`s, so each case is what pydantic-ai
actually resolves: declared, resolved, unresolved, no model, no key.

https://claude.ai/code/session_01SfoF1zJqxTL4WJZadHd5HA


tseaver (changes requested):
Please rebase against `main` (those first two commits from @ggozad are already merged).

src/soliplex/agui/__init__.py
```diff
@@ -271,6 +292,24 @@ async def get_thread(
...
+    ) -> tuple[int, str] | None:
+        """Return the thread's newest measured context size and its run.
+
+        The measurement is 'RunUsage.final_input_tokens' -- the input
+        size of a run's *last* model request, which is what says how
+        full the window was. Runs that never reached the model carry
+        none, and are skipped rather than reported as zero.
+
+        None when no run in the thread has been measured.
+        """
```
tseaver: This query, and the API endpoint which drives it, seem redundant:  we can instead just have the `GET /api/v1/rooms/{room_id}/agui/{thread_id}` endpoint include the `run_usage` for each run, which lets the client loading an existing thread show the correct "window usage" for *each* run in the thread (if desired).

If the client needs a way to fetch the actual final usage for an existing run (the one which it just drove, likely), and doesn't want / need the full payload of `GET /api/v1/rooms/{room_id}/agui/{thread_id}/{run_id}`, then we can add a new `GET /api/v1/rooms/{room_id}/agui/{thread_id}/{run_id}/usage` endpoint which returns it.

In the case where the "last" run has no usage (an error), the client would already have the prior run's usage to fall back on.

src/soliplex/views/agui.py
```diff
@@ -269,6 +269,81 @@ async def get_room_agui_thread_id(
...
+    # measurement still stands -- it was taken from whatever the factory
+    # served -- and a reading with no window hides the gauge.
+    agent_config = room_config.agent_config
+
+    if agent_config.kind == "factory":
+        model_name = None
+        max_model_len = None
+    else:
+        model_name = agent_config.llm_model_name
+        model = config_agents.get_model_from_config(agent_config=agent_config)
```
tseaver: This lookup is for a value which is immutable for a given room across the lifetime of the server process, and should be carried on the `models.DefaultAgent` field the client already gets when it calls `GET /api/v1/rooms/{room_id}`.

If the client uses the `run_usage` from its latest run, or (for a thread the client hasn't yet seen) from the `GET /api/v1/rooms/{room_id}/agui/{thread_id}` query (which we can easily tweak to include `run_usage`), we wouldn't need this endpoint at all.

tseaver (changes requested):

tseaver (approved):
@WilliamKarolDiCioccio thanks, that worked.  One thing I noticed: the new REST endpoint was still paying for loading the full AGUI `RunInput` (including all events), just to return the usage, so I pushed a change adding a new `agui.ThreadStorage.get_run_usage` helper which loads only the usage, after verifying access to the ~~turn~~run.

Issue #1365, opened 2026-09-16
Document / support alembic upgrades to a deployment


tseaver:
Deployments have no documented / supported way to apply the `alembic`
revisions tracked in this repo: released wheels do not include `alembic.ini` or
 `alembic/versions/`.  Today the only path is a source checkout plus
`alembic -x soliplex.installation_path=<path> upgrade head`.

Three cases need covering:

- Existing, never-stamped databases created via `Base.metadata.create_all`
  in `soliplex.installation.lifespan`.  Because that creation does not use
  alembic, no `alembic_version` is written. Such a database is current with
  the schema, but needs stamping at the revision matching the release
  that created the database, before any upgrade.  Covered in #1367, PR #1369.

- New deployments creating their databases need to stamp them with the
  latest alembic version.  Covered in #1368.

- Already-stamped databases. The ordinary `upgrade head`, for both engines
  in the multi-database config (`agui`, `authz`).

Most releases need no migration: there are five revisions so far.  But operators
need a way to check / apply upgrades (or downgrades) following a change
soliplex version pin.  The script in PR #1369 checks for / applies the stamps to
previously unstamped databases.  Later work will handle running normal alembic
upgrades for stamped databases.

Both docs and a `soliplex-cli` wrapper would both help; shipping `alembic.ini`
and `alembic/versions/` as package data would make either usable from a
deployment.

tseaver:
Via #1393.

PR #1366, opened 2026-09-16
feat(agui): report the final reply's tokens and when a run was measured

WilliamKarolDiCioccio:
Follow-up to #1355, prompted by review of the frontend side (soliplex_frontend#545 / #555 / #557). Two gaps in what a run's usage reports, both of which the client was compensating for and neither of which it can fix on its own.

## The reading was one reply short

`final_input_tokens` is the input of the run's last model request. The reply to that request is what the thread now ends with, and the *next* request carries it as input — so the number a client puts over the window is short by exactly one reply, and a draft estimate on top of it can still call the window emptier than it is. That is the same class of gap as cumulative-vs-final in #1355, and the same message that yields the input yields its output. `_final_request_usage` now returns `final_output_tokens` beside it; occupancy is the sum of the two.

**Verified live** against Ollama serving `gpt-oss` on this branch: run 1 measured `95 + 86`, run 2's request then measured `117` in. The reply *was* carried (the input grew), and the count is the provider's own, so on a reasoning model it includes the reasoning — which the next request did not carry (86 reported, ~22 arrived). The sum therefore reads high on such models by the last reply's reasoning, never low, until the next run measures. Ollama itemises no reasoning count, so nothing more exact exists for it; subtracting the itemised counts other providers report would risk reading low on the ones that do send reasoning back, so the raw count is reported and the docstring says so.

## Ordering came from the client's request sequence

`run_usage.created` has always been on the row and never left the database. A client learns of measurements out of order — one run's usage request can resolve after a later run's, and the thread history can land after a run has already reported — and a thread can be run from more than one client. The frontend was reconstructing order from a fetch counter. `measured_at` exposes the timestamp so the newer of two measurements is decided by the backend's clock.

## Change

- `RunUsage.final_output_tokens`, nullable, alembic revision `b7e2f41c9d05` (upgrade, `check` and downgrade verified against a DB stamped at `a1c7d3e90b42`).
- `RunUsageStats` gains `final_output_tokens` and `measured_at` with `None` defaults, so rows written before the column read back as their sibling columns do.
- `AGUI_RunUsage` carries both, on `GET .../agui/{thread_id}`, `GET .../{run_id}` and `GET .../{run_id}/usage`.

No endpoint or config change. 4439 tests, 100% branch coverage.

https://claude.ai/code/session_01SfoF1zJqxTL4WJZadHd5HA


src/soliplex/agui/__init__.py
```diff
@@ -104,10 +104,12 @@ def __init__(
         "tool_calls",
         "final_input_tokens",
         "resolved_model_name",
+        "final_output_tokens",
+        "measured_at",
```
tseaver: I don' understand the `measured_at` field's purpose, and it isn't in the alembic migration:  is it a fossil of an earlier pass?

src/soliplex/agui/__init__.py
```diff
@@ -142,6 +144,30 @@ class RunUsage(abc.ABC):
...
+    thread's occupancy. None when the run produced no model response.
+
+    The provider's own count, as reported: on a reasoning model it includes
+    the reasoning, which providers generally do not carry into the next
+    request, so the sum reads high there by the last reply's reasoning --
+    never low -- until the next run measures. Ollama itemises no reasoning
+    count, so nothing more exact is available for it.
+    """
+
+    measured_at: datetime.datetime | None
```
tseaver: Here as well.  Looking below, it appears to be the timestamp of the end of the run:  why is that needed, given that the client already has that information?

src/soliplex/agui/__init__.py
WilliamKarolDiCioccio: Not a fossil — it was `run_usage.created`, an existing column, exposed through the tuple; hence no migration. But your second comment is right that it is redundant, so it is gone in 6f81360.

src/soliplex/agui/__init__.py
WilliamKarolDiCioccio: Agreed. `AGUI_Run` already carries `created` and `finished` on the listing, and the client fetching `/usage` for a run is the one that drove it. Dropped in 6f81360; the PR is now `final_output_tokens` only.

tseaver:
@WilliamKarolDiCioccio you've unfortunately landed on the work I've been doing to make our database migration story clearer (#1365 and its sub-issues).  Your changes fit the state of `main` from before PR #1373 landed, and will work mostly cleanly for a backport to the `v0.81.x` branch:  see PR #1390.

On `main`, however, the alembic revisions have moved due to work on `main` to make handling them simpler:  they now live in `src/alembic_migrations/`, where the latest head is `a1c7d3e90b42_soliplex_v0_80.py` (should have been renamed, as it landed in `soliplex v0.81`.  Can you get Claude to help you rebase the current branch on `main`?

WilliamKarolDiCioccio:
Rebased onto main. The migration now lives at `src/soliplex/alembic_migrations/versions/b7e2f41c9d05_soliplex_v0_82.py` — same revision id and `down_revision` as the one released in v0.81.1, so a database upgraded through either path lands on the same head. Import grouping follows #1389.

Verified on the new layout: `upgrade a1c7d3e90b42` → no column; `upgrade head` → version `b7e2f41c9d05`, column present; `alembic check` → no new operations; `downgrade a1c7d3e90b42` → column removed. `scripts/lint_alembic_chain.py` reports the chain reproduces the models exactly (agui and authz identical), and its self-test passes. 4565 tests, 100% branch coverage.

The source diff is byte-identical to your backport in #1390.

